PR #12, opened 2021-05-11
Fixed typo error in function name

ShwethaKumbla:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it:
There is a misspelling of function name. This patch fixes it.

#### Which issue(s) this PR fixes:
Fixes #11

PR #14, opened 2021-05-18
[WIP] Enable golanci-lint, add rule for testing and fix lints

cpanato:


cpanato:
@vladimirvivien @spiffxp lets merge this https://github.com/kubernetes-sigs/e2e-framework/pull/12

spiffxp (approved):
/approve
/lgtm
/hold
Remove hold when ready to merge

cpanato:
@spiffxp before merging this i would lile to have the test-infra pr merged then can validate with the changes here :-)

pkg/env/action.go
```diff
@@ -31,21 +31,23 @@ const(
 
 // action a group env functions
 type action struct {
-	role actionRole
-	ctx context.Context
+	role  actionRole
+	ctx   context.Context // nolint
```
vladimirvivien, line 35: If I recall, this is not used right now, field `ctx` can be removed.

vladimirvivien (commented):
@cpanato thanks for this awesome overhaul.
Just had one comment.

pkg/env/action.go
cpanato: yes, this is not used, but i thought is something that will be needed, that why I added the nolint :) but will remove

cpanato: thanks for your Review!

PR #15, opened 2021-05-19
initial commit for klient helper package

ShwethaKumbla:
#### What this PR does / why we need it
Initial implementation of rest.Config constructor functions which helps to get the client connection with the api-server.

### DIFF
--- a/klient/conf/config.go
+++ b/klient/conf/config.go
@@ -1,0 +1,132 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package conf
+
+import (
+	"flag"
+	"os"
+	"os/user"
+	"path"
+
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+// DefaultClusterContext default cluster context
+var DefaultClusterContext = ""
+
+// New returns Kubernetes configuration value of type *rest.Config.
+// filename is kubeconfig file
+func New(fileName string) (*rest.Config, error) {
+	// if filename is not provided assume in-cluster-config
+	if fileName == "" {
+		return rest.InClusterConfig()
+	}
+
+	// create the config object from k8s config path
+	return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: fileName}, &clientcmd.ConfigOverrides{}).ClientConfig()
+}
+
+// NewWithContextName returns k8s config value of type *rest.Config
+func NewWithContextName(fileName, context string) (*rest.Config, error) {
+	// create the config object from k8s config path and context
+	return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: fileName},
+		&clientcmd.ConfigOverrides{
+			CurrentContext: context,
+		}).ClientConfig()
+}
+
+// NewInCluster for clients that expect to be
+// running inside a pod on kubernetes
+func NewInCluster() (*rest.Config, error) {
+	return rest.InClusterConfig()
+}
+
+// ResolveKubeConfigFile returns the kubeconfig file from
+// either flag --kubeconfig or env KUBECONFIG.
+// If flag.Parsed() is true then lookup for --kubeconfig flag.
+// If --kubeconfig, or KUBECONFIG, or  $HOME/.kube/config not provided then
+// assume in cluster.
+func ResolveKubeConfigFile() string {
+	var kubeConfigPath string
+
+	// If a flag --kubeconfig  is specified with the config location, use that
+	if flag.Parsed() {
+		f := flag.Lookup("kubeconfig")
+		if f != nil {
+			return f.Value.String()
+		}
+	}
+
+	// if KUBECONFIG env is defined then use that
+	kubeConfigPath = os.Getenv(clientcmd.RecommendedConfigPathEnvVar)
+	if kubeConfigPath != "" {
+		return kubeConfigPath
+	}
+
+	var (
+		homeDir string
+		ok      bool
+	)
+
+	// check if $HOME/.kube/config is present
+	// if $HOME is unset, get it from current user
+	if homeDir, ok = os.LookupEnv("HOME"); !ok {
+		u, err := user.Current()
+		if err != nil {
+			// consider it as in-cluster-config
+			return ""
+		}
+
+		kubeConfigPath = path.Join(u.HomeDir, clientcmd.RecommendedHomeDir, clientcmd.RecommendedFileName)
+	} else {
+		kubeConfigPath = path.Join(homeDir, clientcmd.RecommendedHomeDir, clientcmd.RecommendedFileName)
+	}
+
+	// check if the config path exists
+	if fileExists(kubeConfigPath) {
+		return kubeConfigPath
+	}
+
+	return ""
+}
+
+// ResolveClusterContext returns cluster context name based on --context flag.
+func ResolveClusterContext() string {
+	// If a flag --kube-context is specified use that
+	if flag.Parsed() {
+		f := flag.Lookup("kube-context")
+		if f != nil {
+			return f.Value.String()
+		}
+	}
+
+	return ""
+}
+
+// fileExists reports whether the named file or directory exists.
+func fileExists(filePath string) bool {
+	if _, err := os.Stat(filePath); err != nil {
+		if os.IsNotExist(err) {
+			return false
+		}
+	}
+
+	return true
+}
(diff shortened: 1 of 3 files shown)


vladimirvivien:
@cpanato I think the CI job may be broken
```
make: *** No rule to make target 'test'.  Stop.
```
Any idea? Your help is appreciated.

cpanato:
> @cpanato I think the CI job may be broken
> 
> ```
> make: *** No rule to make target 'test'.  Stop.
> ```
> 
> Any idea? Your help is appreciated.

@vladimirvivien  we need to merge this: https://github.com/kubernetes-sigs/e2e-framework/pull/14

vladimirvivien:
@cpanato Ha! Just approved #14.
Thanks.

cpanato:
@ShwethaKumbla can you rebase this PR and fix the conflict? thanks!

ShwethaKumbla:
@cpanato done rebase.

ShwethaKumbla:
@cpanato @vladimirvivien 
Do we have kube config file under `/root/.kube/config` in the test env where these jobs are running ?
What should be the ideal way to configure the kube config path in `_test.go` files ?

cpanato:
> @cpanato @vladimirvivien
> Do we have kube config file under `/root/.kube/config` in the test env where these jobs are running ?
> What should be the ideal way to configure the kube config path in `_test.go` files ?

no, it does not have, you might need to create it as part of the test. 
but this is just the unit tests, when we need to test e2e (like enable the examples) we will need to run it on a environment that have a cluster deployed

klient/conf/config.go
```diff
@@ -0,0 +1,124 @@
...
+	"flag"
+	"os"
+	"os/user"
+	"path"
+
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+var DefaultClusterContext = ""
```
vladimirvivien, line 29: ~This var is not needed.~

klient/conf/config.go
```diff
@@ -0,0 +1,124 @@
...
+
+// New returns Kubernetes configuration value of type *rest.Config.
```
vladimirvivien, line 31: Add comment to indicate that the parameter is the kubeconfig file.

klient/conf/config.go
```diff
@@ -0,0 +1,124 @@
...
+func New(fileName string) (*rest.Config, error) {
+	// if filename is not provided assume in-cluster-config
+	if fileName == "" {
```
vladimirvivien, line 34: This logic is OK to assume inClusterConfig if kubeconfig file name is not provided.
We can be explicit by providing a constructor function `NewInCluster()` that specifically assumes in-cluster.

klient/conf/config.go
```diff
@@ -0,0 +1,124 @@
...
+	return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: fileName},
+		&clientcmd.ConfigOverrides{
+			CurrentContext: context,
+		}).ClientConfig()
+}
+
+// ResolveKubeConfigFile returns the kubeconfig file from
+// either flag --kubeconfig or env KUBECONFIG.
+// It will only call flag.Parse() if flag.Parsed() is false.
```
vladimirvivien, line 55: This comment is wrong, the code does not call `flag.Parse()` directly and shouldn't.  As you have in the code, it detects to see if flag.Parsed() is true, then look for the proper flag.

klient/conf/config.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// check if $HOME/.kube/config is present
+	// if $HOME is unset, get it from current user
+	if homeDir, ok = os.LookupEnv("HOME"); !ok {
+		u, err := user.Current()
+		if err != nil {
+			// consider it as in-cluster-config
+			return ""
+		}
+
```
vladimirvivien, line 88: Extra line.

klient/conf/main_test.go
```diff
@@ -0,0 +1,124 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"k8s.io/client-go/util/homedir"
+)
+
+func TestMain(m *testing.M) {
```
vladimirvivien, line 31: Maybe you could use the e2e test-harnes framework to setup your test-main ? 
😆

klient/conf/config_test.go
```diff
@@ -0,0 +1,57 @@
...
+		t.Error("error while client connection", err)
+	}
+
+	if cfg == nil {
+		t.Errorf("client config is nill")
+	}
+}
+
+func TestNewWithContextName(t *testing.T) {
+	cfg, err := NewWithContextName(ResolveKubeConfigFile(), DefaultClusterContext)
```
vladimirvivien, line 49: Ignore my comment about removing `DefaultClusterContext`, I see how you are using it.

vladimirvivien (commented):
Overall looks good.
Letft some comments.

klient/conf/main_test.go
```diff
@@ -0,0 +1,124 @@
...
+	flag.Parse()
+}
+
+func createFile(path, data string) error {
+	return ioutil.WriteFile(path, []byte(data), 0644)
+}
+
+// genKubeconfig used to genearte kube config file
+// we can provide multiple contexts as well
+func genKubeconfig(contexts ...string) string {
```
andrewsykim, line 83: Wondering if we should use the real type representing kubeconfig and json marshal it instead.

andrewsykim: (I realize this is just for tests so maybe not a big deal)

klient/conf/config.go
```diff
@@ -0,0 +1,124 @@
...
+		if f != nil {
+			return f.Value.String()
+		}
+	}
+
+	return ""
+}
+
+// exists reports whether the named file or directory exists.
+func exists(name string) bool {
```
andrewsykim, line 116: Wondering if `path` or `filePath` could be a better name for parameter. Also helps if specify if this expects absolute paths or relative paths

klient/conf/config.go
```diff
@@ -0,0 +1,124 @@
...
+// either flag --kubeconfig or env KUBECONFIG.
+// It will only call flag.Parse() if flag.Parsed() is false.
+// If --kubeconfig, or KUBECONFIG, or  $HOME/.kube/config not provided then
+// assume in cluster.
+func ResolveKubeConfigFile() string {
+	var kubeConfigPath string
+
+	// If a flag --kubeconfig  is specified with the config location, use that
+	if flag.Parsed() {
+		f := flag.Lookup("kubeconfig")
```
andrewsykim, line 63: Should the flag definition of `kubeconfig` be somewhere in this package? I saw it in the test file but wondering if it should be somewhere in this file.
```
flag.StringVar(&kubeconfig, "kubeconfig", "", "Paths to a kubeconfig. Only required if out-of-cluster.")
```

klient/conf/config_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	if filename != filepath.Join(home, "test", ".kube", "config") {
+		t.Errorf("unexpected config path: %s", filename)
+	}
+}
+
+func TestNew(t *testing.T) {
+	cfg, err := New(ResolveKubeConfigFile())
+	if err != nil {
+		t.Error("error while client connection", err)
```
wilsonehusin, line 40:
```suggestion
		t.Error("error while creating client connection", err)
```

klient/conf/main_test.go
```diff
@@ -0,0 +1,124 @@
...
+func setup() {
+	home := homedir.HomeDir()
+
+	path := filepath.Join(home, "test", ".kube", "config")
+
+	// check if file exists
+	_, err := os.Stat(path)
+	// create file if not exists
+	if os.IsNotExist(err) {
+		err = os.MkdirAll(home+"/test/.kube", 0777)
```
wilsonehusin, line 47: 🤔 this is a test so it's probably not that big of a deal, just preventing yet-another-OS-specific implementation that could be easily avoided 😄 

```suggestion
	kubeconfigdir := filepath.Join(home, "test", ".kube")
	kubeconfigpath := filepath.Join(kubeconfigdir, "config")

	// check if file exists
	_, err := os.Stat(kubeconfigpath)
	// create file if not exists
	if os.IsNotExist(err) {
		err = os.MkdirAll(kubeconfigdir, 0777)
```

klient/conf/config.go
vladimirvivien: @andrewsykim I would like to avoid forcing flag parsing in the package. This code assumes package declartion/definition is done elsewhere and only if the flag is defined it will look for it.

klient/conf/config.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	return ""
+}
+
+// ResolveClusterContext returns cluster context name based on --context flag.
+func ResolveClusterContext() string {
+	// If a flag --context is specified use that
+	if flag.Parsed() {
+		f := flag.Lookup("context")
```
andrewsykim, line 106: Wondering if we want to search for something more specific like `--kube-context` or something else. Will this also be supported with env vars?

ShwethaKumbla: This context is for kube context, I will rename this flag as kube-context as it leads to a confusion.  Environment variable is not supported as of now for this.

klient/conf/main_test.go
ShwethaKumbla: @vladimirvivien next PR will handle this.

klient/conf/config.go
```diff
@@ -0,0 +1,132 @@
...
+		if f != nil {
+			return f.Value.String()
+		}
+	}
+
+	return ""
+}
+
+// exists reports whether the named file or directory exists.
+func exists(filePath string) bool {
```
vladimirvivien, line 124: let's be more explicit with that function name, i.e. `fileExists(filepath String) bool`

vladimirvivien (commented):
I only have one comment. Once resolved this is good to go!

vladimirvivien:
+1 on the commit squashing!

vladimirvivien (approved):
LGTM

PR #20, opened 2021-06-14
CRUD operations for managing cluster objects

ShwethaKumbla:
#### What this PR does / why we need it
Initial implementation for Create/Update/Delete/Get/List of cluster resources. This ops helps user to manage the resource efficiently.


go.mod
```diff
@@ -3,6 +3,8 @@ module sigs.k8s.io/e2e-framework
 go 1.16
 
 require (
+	k8s.io/api v0.21.0
 	k8s.io/apimachinery v0.21.0
 	k8s.io/client-go v0.21.0
+	sigs.k8s.io/controller-runtime v0.8.3
```
vladimirvivien, line 9: Now that we're relying on ctrl-runtime, we should make sure we're using latest stable. It looks like it's v0.9.0.

klient/k8s/res/main_test.go
```diff
@@ -0,0 +1,195 @@
...
+func setup() {
+	home := homedir.HomeDir()
+
+	path := filepath.Join(home, "test", ".kube", "config")
+
+	// check if file exists
+	_, err := os.Stat(path)
+	// create file if not exists
+	if os.IsNotExist(err) {
+		err = os.MkdirAll(home+"/test/.kube", 0777)
```
vladimirvivien, line 68: Use filepath.Join() here as well.

vladimirvivien: Well, this is test, does not matter. Ignore previous.

klient/k8s/res/resource.go
```diff
@@ -0,0 +1,117 @@
...
+	res := &Resources{
+		config: cfg,
+		scheme: scheme.Scheme,
+	}
+
+	return res
+}
+
+// New wrapper fumction for controller runtimr client New()
+func (r *Resources) New() cr.Client {
```
vladimirvivien, line 55: This method is leaking a controller-runtime client. is that right ? Probably should think about wrapping it into a klient.Client or something.

vladimirvivien: Actually, better yet, dont make make `New()` visible because all interaction should go through the `Res()` constructor function any way.

klient/k8s/res/resource.go
```diff
@@ -0,0 +1,117 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	cr "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/e2e-framework/klient/k8s"
+)
+
+type Resources struct {
```
vladimirvivien, line 32: Add a `client` field of type `controller-runtime.Client` in this struct so that we create it once and store it here.

klient/k8s/res/resource.go
```diff
@@ -0,0 +1,117 @@
...
+	scheme *runtime.Scheme
+}
+
+func Res(cfg *rest.Config) *Resources {
+	if cfg == nil {
+		fmt.Println("must provide rest.Config")
+		return nil
+	}
+
+	res := &Resources{
```
vladimirvivien, line 46: Initialize the cr.Client here and store it.

klient/k8s/res/resource.go
vladimirvivien: As I mentioned, let's create and store the cr.Client in the Resource struct. Then all this goes away.

klient/k8s/res/resource.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		// TODO: log error and return
+		return nil
+	}
+
+	return client
+}
+
+func (r *Resources) Get(ctx context.Context, name, namespace string, obj k8s.Object) error {
+	return r.New().Get(ctx, cr.ObjectKey{Namespace: namespace, Name: name}, obj)
```
vladimirvivien, line 66: When the `r` (Resources) has the instance of the cr.Client, this should be changed to just used the stored client
```go
r.client.Get(ctx, cr.ObjectKey{Namespace: namespace, Name: name}, obj)
```

vladimirvivien (commented):
This is a great start and I am excited about the possibilities 🎉 
I left some comments. Mainly, let's hide the controller-runtime.Client!
Then we can do another round of reviews soon!

klient/k8s/testing/kindcluster.go
```diff
@@ -0,0 +1,140 @@
...
+	fmt.Println("installing: go get sigs.k8s.io/kind@", kindVersion)
+	p := e.SetEnv("GO111MODULE", "on").RunProc(fmt.Sprintf("go get sigs.k8s.io/kind@%s", kindVersion))
+	if p.Err() != nil {
+		return fmt.Errorf("failed to install kind: %s: %s", p.Err(), p.Result())
+	}
+	if !p.IsSuccess() || p.ExitCode() != 0 {
+		return fmt.Errorf("failed to install kind: %s", p.Result())
+	}
+	return nil
+}
```
cpanato, line 140: maybe is good to add kind in the PATH as well

klient/k8s/testing/kindcluster.go
```diff
@@ -0,0 +1,170 @@
...
+	name string
+	e    *gexe.Echo
+}
+
+func NewKindCluster(name string) *KindCluster {
+	return &KindCluster{name: name, e: gexe.New()}
+}
+
+func (k *KindCluster) Create() error {
+	fmt.Println("Creating kind cluster ", k.name)
```
vladimirvivien, line 40: Consider using log if you are going to keep these print statements.

klient/k8s/testing/kindcluster.go
```diff
@@ -0,0 +1,170 @@
...
+	if ki.Err() != nil {
+		fmt.Println("docker is not present")
+	} else {
+		fmt.Println("docker exist:", ki.Result())
+	}
+
+	fmt.Println("launching: kind create cluster --name", k.name)
+	p := k.e.RunProc(fmt.Sprintf(`kind create cluster --name %s`, k.name))
+	if p.Err() != nil {
+		return fmt.Errorf("failed to create kind cluster: %s", p.Err())
```
vladimirvivien, line 69: To help debug the error, see if the proc returned a message like so
```go
return fmt.Errorf("failed to create kind cluster: %s: %s", p.Err(), p.Result()) 
```

vladimirvivien (commented):
@ShwethaKumbla I see that kinda and docker are able to run OK, but test still fails.  I left a message to help return more message to help with debug.

vladimirvivien (approved):
/approve

1
```diff
@@ -0,0 +1,20 @@
+[WIP] CRUD operations for managing cluster objects
```
cpanato, line 1: @ShwethaKumbla @vladimirvivien this file is needed?

PR #24, opened 2021-07-01
Context propagation fix and context examples

### LINKED ISSUE #21: Context values are not propagated through Before and Setup steps
In the Before and Setup test steps I tried to return a new context with a value based on the parent context. For example:

```
	testenv.Setup(func(ctx context.Context) (context.Context, error) {
	        metadata := getMetadata(..)
	        ctx = context.WithValue(ctx, "some-metadata", metadata)
		return ctx, nil
	})
```

When I then try to access the context value from the Assess step, the value is not available. I was wondering if context values are expected to be propagated throughout the various test steps.

vladimirvivien:
This PR fixes #21. It implements the updated design of how context will work (see [Environment and Context](https://github.com/kubernetes-sigs/e2e-framework/blob/master/docs/design/test-harness-framework.md#envenvironment-and-contextcontext) in doc)

The changes in this PR will include:
* [x] Doc update
* [x] Refactor to update `context.Context` propagation
* [x] New examples


pkg/env/env.go
```diff
@@ -132,9 +132,10 @@ func (e *testEnv) Finish(funcs ...Func) types.Environment {
 func (e *testEnv) Run(ctx context.Context, m *testing.M) int {
 	setups := e.GetSetupActions()
 	// fail fast on setup, upon err exit
+	var err error
 	for _, setup := range setups {
-		// context not surfaced further
-		if _, err := setup.run(ctx); err != nil {
+		// context passed down to each setup
+		if ctx, err = setup.run(ctx); err != nil {
```
andrewsykim, line 138: Is this returned context being passed into Assess somewhere?

vladimirvivien: @andrewsykim I realize what the issue is now.
No, no way to pass anything into running tests using testing.M
So, let me tinker with this to see if there is a better way (other than global context shared).

vladimirvivien:
@andrewsykim Changing this to work in progress as the fix may be longer than I thought.
Initially, the Setup and Finish test env methods were not designed to propagate the context because of the limitation of the testing.M.  Meaning, the test author would initialize a context in an main_test.go file (responsible for initializing test suite).  Then, globally (package scope) propagate that context to actual tests.

Admittedly, because Go's testing framework does not support Context, it makes this approach above kinda awkward. A better approach would be to create a context at env.Setup, then (intentionally or automatically) have that context be propagated downstream.  I think it's doable, but may change how tests are executed and change the design a few places.

PR #26, opened 2021-07-06
Adds tests for k8s resource

tech-geek29:
This PR adds tests for resource.go. Following tests are added:
1. Test for instantiating the controller runtime client without Config
2. Test for instantiating the controller runtime client with invalid Config
3. Test for Update Resource
4. Test for Delete Resource

/cc @ShwethaKumbla


cpanato (commented):
/ok-to-test

cpanato (approved):
thanks!
/lgtm
/assign @vladimirvivien

ShwethaKumbla:
Thanks @tech-geek29.
/lgtm

ShwethaKumbla (approved):

klient/k8s/res/resource_test.go
```diff
@@ -80,6 +81,84 @@ func TestRes(t *testing.T) {
...
 }
 
+func TestResNoConfig(t *testing.T) {
+	defer func() {
+		if r := recover(); r == nil {
+			t.Error("expected panic while invoking Res without k8s config")
+		}
+	}()
+
+	Res(nil)
```
andrewsykim, line 91: Random question, how come Res panics instead of returning error?

tech-geek29: I am also not very sure but I believe if controller runtime client is not instantiated then the normal execution of goroutine should be stopped as there is nothing that can be done without Resource object.

tech-geek29: Also if we consider returning error, then also the option that is left with calling method is check error is exit (as it cannot proceed with normal flow, since that requires resource)

klient/k8s/res/resource_test.go
```diff
@@ -80,6 +81,84 @@ func TestRes(t *testing.T) {
...
+		if r := recover(); r == nil {
+			t.Error("expected panic while invoking Res with invalid k8s config")
+		}
+	}()
+
+	cfg := &rest.Config{
+		Host: "invalid-host",
+	}
+
+	Res(cfg)
```
andrewsykim, line 105: Not related to this specific PR but it would be good to rename this to use Resource and use the full term going forward (like in the package name too). "Res" can be interpreted as a lot of things and I don't think the shorthand form is that much easier than just "Resource"

tech-geek29: I agree on this! Will have a separate PR for this change.

klient/k8s/res/resource_test.go
ShwethaKumbla: @tech-geek29 Yes, you are correct. If controller runtime client is not instantiated then execution should be stopped as there is nothing can be done without Resource. So we have added panic and not error handling as this will be a rare scenario.

klient/k8s/res/resource_test.go
tech-geek29: @andrewsykim I will be raising a separate PR for this change.

andrewsykim (commented):
/approve

PR #27, opened 2021-07-07
Rename Res method to Resource

tech-geek29:
This PR renames Res method under klient/k8s to Resource.

This was suggested by @andrewsykim as part of review comments [here](https://github.com/kubernetes-sigs/e2e-framework/pull/26#discussion_r664695191)

/cc @ShwethaKumbla

### DIFF
--- a/klient/k8s/resources/main_test.go
+++ b/klient/k8s/resources/main_test.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package res
+package resources
 
 import (
 	"context"
--- a/klient/k8s/resources/resources.go
+++ b/klient/k8s/resources/resources.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package res
+package resources
 
 import (
 	"context"
@@ -41,21 +41,22 @@ type Resources struct {
 	client cr.Client
 }
 
-// Res instantiates the controller runtime client
+// New instantiates the controller runtime client
 // object. User can get panic for belopw scenarios.
 // 1. if user does not provide k8s config
 // 2. if controller runtime client instantiation fails.
-func Res(cfg *rest.Config) *Resources {
+func New(cfg *rest.Config) (*Resources, error) {
 	if cfg == nil {
 		// TODO: logging
 		fmt.Println("must provide rest.Config")
-		panic(errors.New("must provide rest.Config"))
+		return nil, errors.New("must provide rest.Config")
 	}
 
 	cl, err := cr.New(cfg, cr.Options{Scheme: scheme.Scheme})
 	if err != nil {
 		// TODO: log error
-		panic(err)
+		fmt.Println("unexpected error creating client using provided config and client options")
+		return nil, err
 	}
 
 	res := &Resources{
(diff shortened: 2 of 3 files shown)


cpanato (commented):
/ok-to-test

klient/k8s/res/resource.go
```diff
@@ -41,11 +41,11 @@ type Resources struct {
 	client cr.Client
 }
 
-// Res instantiates the controller runtime client
+// Resource instantiates the controller runtime client
 // object. User can get panic for belopw scenarios.
 // 1. if user does not provide k8s config
 // 2. if controller runtime client instantiation fails.
-func Res(cfg *rest.Config) *Resources {
```
andrewsykim, line 48: This is a constructor, so it should probably be named something like `NewResources()`.

andrewsykim (commented):
Can we also rename the package?

klient/k8s/res/resource.go
tech-geek29: I missed package renaming. Thanks! 
I have renamed package to resource. Also, I have changed the constructor to `NewResource`

klient/k8s/resource/resource.go
```diff
@@ -41,11 +41,11 @@ type Resources struct {
 	client cr.Client
 }
 
-// Res instantiates the controller runtime client
+// NewResource instantiates the controller runtime client
 // object. User can get panic for belopw scenarios.
 // 1. if user does not provide k8s config
 // 2. if controller runtime client instantiation fails.
-func Res(cfg *rest.Config) *Resources {
+func NewResource(cfg *rest.Config) *Resources {
```
andrewsykim, line 48: NewResources because the type is named Resources? Same with package naming, should it be `resources`?  Maybe for another follow-up

klient/k8s/resource/resource.go
```diff
@@ -41,11 +41,11 @@ type Resources struct {
 	client cr.Client
 }
 
-// Res instantiates the controller runtime client
+// NewResource instantiates the controller runtime client
 // object. User can get panic for belopw scenarios.
 // 1. if user does not provide k8s config
```
andrewsykim, line 46: As a follow-up, this should probably return an error and the tests can be updated to check the error instead of capturing a panic

klient/k8s/resource/resource.go
tech-geek29: I think the struct type should be renamed to Resource. Any reason to name type as Resources? In my opinion slices/array type names should be plural, but struct represents a single user-defined type.

klient/k8s/resource/resource.go
tech-geek29: I think if initialisation of controller runtime client fails, the flow should be terminated as without client there is no operation that can be performed!

klient/k8s/resource/resource.go
vladimirvivien: I wanted to do a fluent style API and suggested the panic.
I am ok if we change it to return an error and let callee decide what to do with error.

vladimirvivien (commented):
Thanks for looking into this.

klient/k8s/resource/resource.go
tech-geek29: Renamed `NewResource` to `New` as package name is resources. Also removed panic from `New` instead made it to return an error and let caller decided how to handle the error.

vladimirvivien (approved):
LGTM

PR #33, opened 2021-07-14
helper functions for handling k8s objects

ShwethaKumbla:
This PR adds additional helper methods to handle k8s objects.

### DIFF
--- a/klient/k8s/resources/resources.go
+++ b/klient/k8s/resources/resources.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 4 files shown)


klient/k8s/resources/resources.go
```diff
@@ -121,3 +133,49 @@ func (r *Resources) List(ctx context.Context, objs k8s.ObjectList, opts ...ListO
...
+	p := cr.RawPatch(patch.PatchType, patch.Data)
+
+	o := &cr.PatchOptions{Raw: patchOptions}
+	return r.client.Patch(ctx, objs, p, o)
+}
+
+// BuildFromJSON encodes a JSON string as a typed (or unstructured) object representation.
+// TODO: needs to define the object structure to full fill the functionality
+func (r *Resources) BuildFromJSON(obj k8s.Object, jsonData string) error {
+	return json.Unmarshal([]byte(jsonData), &obj)
```
vladimirvivien, line 170: I don't think we should use the plain JSON encoder.
Instead, look in the [api-machinery runtime for tests on how to use encoders](https://sourcegraph.com/github.com/kubernetes/apimachinery@master/-/blob/pkg/runtime/embedded_test.go)

klient/k8s/resources/resources.go
```diff
@@ -121,3 +133,49 @@ func (r *Resources) List(ctx context.Context, objs k8s.ObjectList, opts ...ListO
 	o := &cr.ListOptions{Raw: listOptions}
 	return r.client.List(ctx, objs, o)
 }
+
+func WithLabelSelector(sel string) ListOption {
```
vladimirvivien, line 137: Love seeing these convenience functions for the options!!!

klient/k8s/resources/resources.go
vladimirvivien: Actually this package is better and it applies the proper object conversions https://sourcegraph.com/github.com/kubernetes/apimachinery@master/-/tree/pkg/util/json

vladimirvivien (commented):
This is awesome work. I left comment on using the runtime json converters instead of plain json encoder. See https://sourcegraph.com/github.com/kubernetes/apimachinery@master/-/tree/pkg/util/json

klient/k8s/resources/resources.go
ShwethaKumbla: Just marked this PR to WIP to update the changes for JSON encoding .

ShwethaKumbla: JSON encoding will be taken care as part of #43

vladimirvivien (approved):
lgtm

Issue #32, opened 2021-07-12
Simplify how Environment is created


vladimirvivien:
Now that the `klient` package has landed, the way that an `environment` is created should be revisited.  The environment type should use a `klient` value to create/keep track of internal *rest.Client value.

* [ ] Update doc with new approach
* [ ] Implement changes

vladimirvivien:
Fixed by #24

Issue #31, opened 2021-07-09
Support namespace creation/deletion per test case


andrewsykim:
The upstream Kubernetse e2e tests follow the pattern of using a dedicated namespace per test case. This helps with resource clean up and isolation. It would be great if the e2e framework provided this mechanism automatically or via opt-in. Perhaps a single testenv or a feature of a given testenv provides an isolated namespace with a generated name.

vladimirvivien:
I will update the design doc to reflect this #30

PR #34, opened 2021-07-15
Refactor type `Environment` and env `Config` to support  new type`klient.Client`

vladimirvivien:
This PR is needed to refactor the way type `Environment` is constructed and initialized with package `klient` which did not exist when this package was created.  This work is also a precursor to the refactoring work that will take place to fix context propagation.

This potentially will address

*  #31 
*  #32

Context propagation will be handled in a future PR


vladimirvivien:
@cpanato @ShwethaKumbla  @andrewsykim  PTAL

klient/client.go
```diff
@@ -0,0 +1,61 @@
...
+	return New(cfg)
+}
+
+func (c *client) RESTConfig() *rest.Config {
+	return c.cfg
+}
+
+func (c *client) Resources() *resources.Resources {
+	return c.resources
+}
```
cpanato, line 61: nit: new line

pkg/envconf/config.go
```diff
@@ -0,0 +1,153 @@
...
+// Labels returns the environment's label filters
+func (c *Config) Labels() map[string]string {
+	return c.labels
+}
+
+func randNS() string {
+	bytes := make([]byte, 25)
+	rand.Read(bytes)
+	return fmt.Sprintf("testns-%s", hex.EncodeToString(bytes))
+}
```
cpanato, line 153: nit: new line

cpanato (commented):
thanks, lgtm, just need to fix the lint issues we I think we are good to go

pkg/envconf/config.go
```diff
@@ -0,0 +1,153 @@
...
+	return c
+}
+
+// Labels returns the environment's label filters
+func (c *Config) Labels() map[string]string {
+	return c.labels
+}
+
+func randNS() string {
+	bytes := make([]byte, 25)
```
cpanato, line 150: maybe we can use `github.com/google/uuid` and generate an uuid and remove the `testns-%s`.

just an idea :)

cpanato: make sense, as i said just a suggestion :)

PR #38, opened 2021-07-29
add support functionality for kind

ShwethaKumbla:
This PR introduces support helper for kind cluster creation.

### DIFF
--- a/klient/k8s/resources/main_test.go
+++ b/klient/k8s/resources/main_test.go
@@ -32,7 +32,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/util/homedir"
 	"sigs.k8s.io/e2e-framework/klient/conf"
-	test "sigs.k8s.io/e2e-framework/klient/k8s/testing"
+	"sigs.k8s.io/e2e-framework/support/kind"
 )
 
 var (
@@ -44,7 +44,7 @@ var (
 	ctx                = context.TODO()
 	cfg          *rest.Config
 	namespace    *corev1.Namespace
-	kind         *test.KindCluster
+	kc           *kind.Cluster
 )
 
 func TestMain(m *testing.M) {
@@ -59,7 +59,7 @@ func setup() {
 	home := homedir.HomeDir()
 	path := filepath.Join(home, ".kube", "config")
 
-	// setup a kind cluster
+	// set up kind cluster
 	err := setupKindCluster()
 	if err != nil {
 		fmt.Println("error while setting up kind cluster", err)
@@ -92,8 +92,8 @@ func setup() {
 
 // setupKindCluster
 func setupKindCluster() error {
-	kind = test.NewKindCluster("e2e-test-cluster")
-	if err := kind.Create(); err != nil {
+	kc = kind.NewKindCluster("e2e-test-cluster")
+	if _, err := kc.Create(); err != nil {
 		return err
 	}
 	fmt.Println("kind cluster created")
@@ -111,7 +111,7 @@ func teardown() {
 	deleteNamespace(ctx, namespace)
 
 	// delete kind cluster
-	err := kind.Destroy()
+	err := kc.Destroy()
 	if err != nil {
 		fmt.Println("error while deleting the cluster", err)
 		return
--- a/klient/k8s/resources/main_test.go
+++ b/klient/k8s/resources/main_test.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 	"testing"
@@ -62,7 +63,7 @@ func setup() {
 	// set up kind cluster
 	err := setupKindCluster()
 	if err != nil {
-		fmt.Println("error while setting up kind cluster", err)
+		log.Println("error while setting up kind cluster", err)
 		return
 	}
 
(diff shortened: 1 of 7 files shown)


pkg/envsupport/support.go
```diff
@@ -0,0 +1,78 @@
...
+
+func CreateCluster(clusterName string) env.Func {
+	return func(ctx context.Context, cfg *envconf.Config) (context.Context, error) {
+		k := kind.NewKindCluster(clusterName)
+		kubecfg, err := k.Create()
+		if err != nil {
+			return ctx, err
+		}
+
+		fmt.Println("kind cluster created")
```
vladimirvivien, line 59: remove Println

vladimirvivien: Actually remove all fmt. Print statements

pkg/envsupport/support.go
```diff
@@ -0,0 +1,78 @@
...
+
+		// stall to wait for kind pods initialization
+		waitTime := time.Second * 10
+		fmt.Println("waiting for kind pods to initialize...", waitTime)
+		time.Sleep(waitTime)
+		return context.WithValue(context.WithValue(ctx, ContextKeyKubeConfig, kubecfg), ContextKeyClusterName, clusterName), nil
+	}
+}
+
+func DestroyCluster(clusterName string) {
```
vladimirvivien, line 69: If KindCluster.Create() creates a kindconfig file, we should provide a way to delete the kubeconfig file during destroy as well.
May be extract the file name from the context then delete it.

support/kind/kind.go
```diff
@@ -28,51 +29,68 @@ import (
...
+	defer file.Close()
+	if n, err := io.Copy(file, p.Out()); n == 0 || err != nil {
+		return "", fmt.Errorf("kind kubecfg file: bytes copied: %d: %w]", n, err)
+	}
+
+	return file.Name(), nil
 }
 
-func (k *KindCluster) GetKubeConfig() (io.Reader, error) {
+func (k *Cluster) GetKubeConfig() (io.Reader, error) {
```
vladimirvivien, line 83: If you gonna use the code starting on line 65. Then you don't need that method, no ?

vladimirvivien (commented):
This is a great start. Cant wait to use it in the examples.

support/kind/kind.go
ShwethaKumbla: Yes, this function is not required.

pkg/envsupport/support.go
ShwethaKumbla: Yes, missed this cleanup. Will update the PR

klient/k8s/resources/main_test.go
```diff
@@ -92,8 +92,8 @@ func setup() {
 
 // setupKindCluster
 func setupKindCluster() error {
-	kind = test.NewKindCluster("e2e-test-cluster")
-	if err := kind.Create(); err != nil {
+	kc = kind.NewKindCluster("e2e-test-cluster")
```
vladimirvivien, line 95: Function Probably can be named just `NewCluster`

klient/k8s/resources/main_test.go
```diff
@@ -92,8 +92,8 @@ func setup() {
...
+	if _, err := kc.Create(); err != nil {
 		return err
 	}
 	fmt.Println("kind cluster created")
```
vladimirvivien, line 99: Don't forget to remove the `fmt.Print` statements.

pkg/envsupport/support.go
```diff
@@ -0,0 +1,75 @@
...
+	ContextKeyKubeConfig = ContextKey("kubeconfig")
+	// ContextKeyClusterName key for setting clustername value in context
+	ContextKeyClusterName = ContextKey("clustername")
+)
+
+// GetValueFromContext gets the caller value from the context.
+// Use this function to retrieve value of a context key
+// or can print out value of key by doing,
+// fmt.Println("Key is:", ContextKeyKubeConfig.String())
+func GetValueFromContext(ctx context.Context, key ContextKey) (string, bool) {
```
vladimirvivien, line 46: If we're going to provide a helper func to retrieve data in context, we cant assume that the value will be a string. This should be something like 
```go
func GetValueFromContext(ctx context.Context, key ContextKey) (interface{}, bool)
```
Or, rename it to indicate the type returned
```go
func GetStringValueFromContext(ctx, key ContextKey)(string, bool)
```

support/kind/kind.go
```diff
@@ -28,78 +29,65 @@ import (
 
 var kindVersion = "v0.11.0"
```
vladimirvivien, line 30: Let's make this a value that can be changed.  Hardcoding it means the user will be stuck with that version.
So, maybe a new constructor function like
```go
func NewClusterWithVersion(clusterName, version string){}
```

Or, add a method on type `Cluster`

```go
type Cluster struct
   version string
}

func (c *Cluster) WithVersion(ver string) *Cluster {
    c.version = ver
    return c
}
```

vladimirvivien: Also, `Cluster.version` should be initialized with the default version (i.e. `v0.11.0`) in case it's never set using the above.

vladimirvivien (commented):
Left some more comments.
But looking better!!!

vladimirvivien (approved):
/lgtm
/approve

cpanato (approved):
/lgtm

PR #52, opened 2021-09-14
Run only filtered features by their assigned labels using --labels

ShwethaKumbla:
User can run tests by filtering features by their labels using `--labels`. This will run only those tests which matches with the pair of labels passed in option `--labels`


cpanato (approved):
/lgtm

thanks!

pkg/env/env.go
```diff
@@ -286,13 +287,21 @@ func (e *testEnv) getFinishActions() []action {
...
 	// feature-level subtest
 	t.Run(featName, func(t *testing.T) {
+		// skip if feature name does not match
 		if e.cfg.FeatureRegex() != nil && !e.cfg.FeatureRegex().MatchString(featName) {
 			t.Skipf(`Skipping feature "%s": name not matched`, featName)
 		}
 
+		// skip if labels does not match
+		// run tests if --labels values matches the feature labels
+		if e.cfg.Labels() != nil && !reflect.DeepEqual(labels, e.cfg.Labels()) {
```
vladimirvivien, line 301: Deep equality will not work here because it requires the maps (compared) to be of the same length. That will not always be the case using the framework. For instance I may have a set of labels with 3 keys defined in a feature test. However, I only use a map with one entry with --labels argument.

The best thing to do is to loop through the defined labels and look for the provided label.

vladimirvivien (commented):
Thanks for starting this @ShwethaKumbla 
Left a comment.

pkg/env/env.go
ShwethaKumbla: I have updated this check by looping through the defined labels.

Issue #42, opened 2021-08-20
Avoid sharing *testing.T during feature execution context


vladimirvivien:
Currently when `Environment.Test(...)` is called from with a test function as shown below:
```go
func TestSomething(t *testing.T) {
    feature.New().Assess(t, ...)
    env.Test(t, ...)
}
```
Only one instance of `t` passed around from feature to feature.  This can cause issues such as early termination of feature test.  A better approach is to create a new `*testing.T` for each feature. That way if a feature fails, the rest of the feature execution continues (within the same Test function).

PR #55, opened 2021-09-17
Support for table-driven test representation

### LINKED ISSUE #89: Table-driven tests
While creating example tests, I came to the realization that using table-driven tests would be useful to be used with the framework. While the framework provides a nice and structured way of encapsulating assessments, it turns out that a table-driven versions of would be more useful when testing multiple cases of the same functionality.

As a test writer, I should be able to represent my assessments as a series of tests that are driven using Go's table-driven test convention.

## Design
The framework should provide a type that allows test writers to capture assessments in a repeatable structure as follows:

```go
type Table []struct {
	Name        string
	Description string
	Labels      Labels
	Assessment  Func
}
```
Type `Table` should also include a method that automatically gathers the assessment data and construct the proper e2e feature tests as shown below:

```go
env.Test(t, table.Features()...) 
```

## Example

```go
func TestMain(m *testing.M){
	test.Setup(func(ctx context.Context, config *envconf.Config) (context.Context, error) {
		rnd := rand.New(rand.NewSource(time.Now().UnixNano()))
		return context.WithValue(context.WithValue(ctx, "limit", rand.Int31n(255)), "randsrc", rnd), nil
	})

	os.Exit(test.Run(m))
}

func TestTableDriven(t *testing.T) {
	tests := features.Table{
		{
			Name: "less than equal 64",
			Assessment: func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
				rnd := ctx.Value("randsrc").(*rand.Rand)  // in real test, check asserted type
				lim := ctx.Value("limit").(int32) // check type assertion
				if rnd.Int31n(int32(lim)) > 64 {
					t.Error("limit should be less than 64")
				}
				return ctx
			},
		},
		{
			Name: "more than than equal 128",
			Assessment: func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
				rnd := ctx.Value("randsrc").(*rand.Rand)  // in real test, check asserted type
				lim := ctx.Value("limit").(int32) // check type assertion
				if rnd.Int31n(int32(lim)) > 128 {
					t.Error("limit should be less than 128")
				}
				return ctx
			},
		},
	}

	test.Test(t, tests.Features()...)
}
```

vladimirvivien:
This patch introduces support for table-driven tests using the framework.

Fixes #89

### DIFF
--- a/examples/table/table_test.go
+++ b/examples/table/table_test.go
@@ -1,0 +1,95 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package table
+
+import (
+	"context"
+	"math/rand"
+	"os"
+	"testing"
+	"time"
+
+	"sigs.k8s.io/e2e-framework/pkg/env"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+var test = env.New()
+func TestMain(m *testing.M){
+	// Setup the rand number source and a limit
+	test.Setup(func(ctx context.Context, config *envconf.Config) (context.Context, error) {
+		rnd := rand.New(rand.NewSource(time.Now().UnixNano()))
+		return context.WithValue(context.WithValue(ctx, "limit", rand.Int31n(255)), "randsrc", rnd), nil
+	})
+
+	// Don't forget to launch the package test
+	os.Exit(test.Run(m))
+}
+
+func TestTableDriven(t *testing.T) {
+	// feature 1
+	table0 := features.Table{
+		{
+			Name: "less than equal 64",
+			Assessment: func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
+				rnd := ctx.Value("randsrc").(*rand.Rand)  // in real test, check asserted type
+				lim := ctx.Value("limit").(int32) // check type assertion
+				if rnd.Int31n(lim) > 64 {
+					t.Log("limit should be less than 64")
+				}
+				return ctx
+			},
+		},
+		{
+			Name: "more than than equal 128",
+			Assessment: func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
+				rnd := ctx.Value("randsrc").(*rand.Rand)  // in real test, check asserted type
+				lim := ctx.Value("limit").(int32) // check type assertion
+				if rnd.Int31n(lim) > 128 {
+					t.Log("limit should be less than 128")
+				}
+				return ctx
+			},
+		},
+		{
+			Assessment: func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
+				rnd := ctx.Value("randsrc").(*rand.Rand)  // in real test, check asserted type
+				lim := ctx.Value("limit").(int32) // check type assertion
+				if rnd.Int31n(lim) > 256 {
+					t.Log("limit should be less than 256")
+				}
+				return ctx
+			},
+		},
+	}.Build("Random numbers").Feature()
+
+	// feature 2
+	table1 := features.Table{
+		{
+			Name: "A simple feature",
+			Assessment: func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
+				rnd := ctx.Value("randsrc").(*rand.Rand)
+				if rnd.Int() > 100 {
+					t.Log("this is a great number")
+				}
+				return ctx
+			},
+		},
+	}
+
+	test.Test(t, table0, table1.Build().Feature() )
+}
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -21,7 +21,9 @@ package env
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"testing"
+	"time"
 
 	log "k8s.io/klog/v2"
 
@@ -42,6 +44,7 @@ type testEnv struct {
 	ctx     context.Context
 	cfg     *envconf.Config
 	actions []action
+	rnd     rand.Source
 }
 
 // New creates a test environment with no config attached.
@@ -89,6 +92,7 @@ func newTestEnv() *testEnv {
 	return &testEnv{
 		ctx: context.Background(),
 		cfg: envconf.New(),
+		rnd: rand.NewSource(time.Now().UnixNano()),
 	}
 }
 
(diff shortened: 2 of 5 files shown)


vladimirvivien:
@ShwethaKumbla @crandles @harshanarayana @helayoty 

Please take a look let me know if this makes sense. Thanks!

pkg/features/table.go
```diff
@@ -0,0 +1,44 @@
...
+*/
+
+package features
+
+// Table structure for table-driven tests
+type Table []struct {
+	Name        string
+	Description string
+	Labels      Labels
+	Assessment  Func
```
harshanarayana, line 24: With this mode every test in the table is being turned into a feature. Would it make sense to retain the ability of being able to run multiple assessment for the feature intact in this mode as well ?

pkg/features/table.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Features returns a slice of test features to be tested
+// and e2e test environment
+func (table Table) Features() (result []Feature) {
+	for _, test := range table {
+		if test.Description == ""{
```
harshanarayana, line 31: Think this will fail the `gofmt` tests ?

harshanarayana (commented):
Love the ability of being able to run tabled tests. I added a small suggestion. Other than that, rest all look good.

pkg/features/table.go
vladimirvivien: @harshanarayana  That is a good point and I agree. The table represents a collection of assessment around the same feature.

helayoty: How about using Steps to implement the table-driven instead of introducing a new type (`tables`)?

crandles: One benefit to table over feature is the ability to skip naming the table

vladimirvivien: @helayoty You and @harshanarayana are correct, I will update the code to translate each table entry as a step (assessment) in the same feature.  I introduced a new type (table) to avoid overloading type `Step` ... the name table, in context of Go test, has special meaning and that's why I selected it.  But, definitely great input and will revisit it for sure.

Also there is an example already with this PR that shows how to use the

helayoty (approved):
lgtm. This sounds simpler and on spot. Thanks @vladimirvivien

pkg/features/table.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Build converts the defined test steps in the table
+// into a FeatureBuilder which can be used to add additional attributes
+// to the feature before it's exercised.
+func (table Table) Build(name string) *FeatureBuilder {
+	f := New(name)
+	for i, test := range table {
+		if test.Name == "" {
+			test.Name = fmt.Sprintf("%s-%d", test.Name, i)
```
crandles, line 37: Is this what you intend for? (test.Name is "" here)
```suggestion
			test.Name = fmt.Sprintf("%s-%d", tname, i)
```

pkg/features/table.go
```diff
@@ -0,0 +1,44 @@
...
+// Each entry in the table represents an executable assessment.
+type Table []struct {
+	Name       string
+	Assessment Func
+}
+
+// Build converts the defined test steps in the table
+// into a FeatureBuilder which can be used to add additional attributes
+// to the feature before it's exercised.
+func (table Table) Build(name string) *FeatureBuilder {
```
crandles, line 33: Can name be empty?

crandles: Looks like it is used here: https://github.com/kubernetes-sigs/e2e-framework/blob/84123945caf59b33f30ce1cd8f1d9e01a0d3c93b/pkg/env/env.go#L310

and using an empty string does not provide what I was looking for (which is avoiding this sub-test layer with this extra name)

examples/table/table_test.go
```diff
@@ -0,0 +1,70 @@
...
+			Assessment: func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
+				rnd := ctx.Value("randsrc").(*rand.Rand)  // in real test, check asserted type
+				lim := ctx.Value("limit").(int32) // check type assertion
+				if rnd.Int31n(lim) > 128 {
+					t.Error("limit should be less than 128")
+				}
+				return ctx
+			},
+		},
+	}.Build("Random numbers").Feature()
```
crandles, line 67: If you wanted to add a setup/teardown, you would need to add it after building the table, which might be awkward

pkg/features/table.go
```diff
@@ -0,0 +1,44 @@
...
+
+package features
+
+import (
+	"fmt"
+)
+
+// Table provides a structure for table-driven tests.
+// Each entry in the table represents an executable assessment.
+type Table []struct {
```
crandles, line 25: if this satisfied the `Feature` interface, you might not need to have to call Build explicitly

pkg/features/table.go
vladimirvivien: I debated around this myself. In a table-driven setup, the top test should be named automatically.
The issue with the extra-subtest layer is another one that I want to address eventually (#42).
Let me take another pass at it and see what I come up with.

vladimirvivien:
@crandles PTAL.

pkg/env/env.go
```diff
@@ -306,7 +310,9 @@ func (e *testEnv) getFinishActions() []action {
 
 func (e *testEnv) execFeature(ctx context.Context, t *testing.T, f types.Feature) context.Context {
 	featName := f.Name()
-
+	if featName == "" {
+		featName = fmt.Sprintf("%s-feature-%x", t.Name(), e.rnd.Int63())
```
crandles, line 314: The example is resulting in an output that looks like this:

```
TestTableDriven/TestTableDriven-feature-8e14a2df72a380a/A_simple_feature
```

I would drop the `t.Name()` to save a few repeated characters from the output:

```suggestion
		featName = fmt.Sprintf("feature-%x",  e.rnd.Int63())
```

pkg/env/env.go
```diff
@@ -343,8 +349,12 @@ func (e *testEnv) execFeature(ctx context.Context, t *testing.T, f types.Feature
 		// assessments run as feature/assessment sub level
 		assessments := features.GetStepsByLevel(f.Steps(), types.LevelAssess)
 
-		for _, assess := range assessments {
-			t.Run(assess.Name(), func(t *testing.T) {
+		for i, assess := range assessments {
+			assessName := assess.Name()
+			if assessName == "" {
+				assessName = fmt.Sprintf("%s-%d", t.Name(), i)
```
crandles, line 355: Similarly, I would drop `t.Name` here and just use `i` to avoid repetition.
```suggestion
				assessName = fmt.Sprintf("%d", i)
```
or some prefix:

```suggestion
				assessName = fmt.Sprintf("assess-%d",i)
```

pkg/env/env.go
vladimirvivien: Good idea.

pkg/env/env.go
vladimirvivien: Likewise, great suggestion.

vladimirvivien:
@crandles I think i finally got the naming for anonymous feature/assessment names correct.

PTAL

crandles:
This LGTM -- I have one suggestion as far as making the Table one step easier to use: https://github.com/vladimirvivien/e2e-framework/pull/1

vladimirvivien:
@crandles your suggestion is interesting (and would make it easier), but presents couple of issues

* The base type for `Table` is a slice of structs, to keep the table-driven form, and has no where store states (i.e. feature name, labels, etc)
* Because of above, method `Build` returns `FeatureBuilder` simply to capture additional attributes like labels, name, etc.

I can't think of a way of doing this without making the `table` form more complex .

PR #57, opened 2021-09-17
Investigating issues with flags

### LINKED ISSUE #56: Flags support seems broken when test binary generated
The framework is designed to work with or without flags. As such flags are not parsed early to avoid forcing test writers to deal with unnecessary CLI flag parsing.  However, it seems that flags are broken when binary is generated.

```
go test -c -o test.test .
```

When the binary is executed, it generated an error:

```
./test.test --kubeconfig /Users/vivienv/.kube/config
 no configuration has been provided, try setting KUBERNETES_MASTER environment variable
```

vladimirvivien:
Fixes #56


vladimirvivien:
@ShwethaKumbla - this is finally building. PTAL.

Issue #58, opened 2021-10-08
New env should default to in-cluster config


johnSchnake:
When doing a simple API call example I had to do the following:
```
	c,err := envconf.NewWithKubeconfig("")
	if err !=nil{
		t.Fatalf("Failed to get in-cluster config: %v", err)
	}
	e := env.NewWithConfig(c)
```
I would have expected `e := env.New()` to use a default in-cluster config or something but I got a panic.

If I did something wrong let me know, but otherwise I think a tweak should be made to avoid those 4 extra lines of boilerplate, in-cluster-config setup.

PR #54, opened 2021-09-17
Additional test examples

vladimirvivien:
This patch restructured and introduces the kind example.
It also introduces example/lifecycle with detail about
the stages of a test and how to override them.

PR #63, opened 2021-10-11
Make it easier to use in-cluster configurations

johnSchnake:
- New method to create an env with an in-cluster config
- New method to create a Config with an in-cluster config
- New method MustClient(t) to return the client without worrying
about error handling every time.
- New example for simple_API tests that utilize an in-cluster config

Partially addresses: #54 
Fixes: #58 


johnSchnake:
I removed a section of code that was deleting the kubeconfig entirely as I suspected that was causing an issue. Locally I was able to repro the issue and this did fix it; now its still happening in CI. TBD.

johnSchnake:
I don't expect the next set of tests to pass but I'm trying to figure out whats going on in the test invocations on the host. Just filed a few other issues #67 #66

johnSchnake:
After updating to gexe 0.1.1, got a green test run. Trying again to ensure its not a fluke. See https://github.com/vladimirvivien/gexe/issues/41

PR #64, opened 2021-10-11
Add more info for before/after hooks

johnSchnake:
- Adds two different types of functions for actions to store/call
which take features and *testing.T objects
- This allows before/after hooks that can actually reference the
object in question
- This enables an extremely useful and common test case of creating
a namespace for each test
- The namespace-per-test use case is shown in a new example


johnSchnake:
Waiting for #65 to simplify some fo the tests so its debuggable.

johnSchnake:
Waiting for #65 to simplify some of the tests so its debuggable (local branch has some unexpected errs)

PR #71, opened 2021-11-02
Rework the FeatureInfo wiring

### LINKED ISSUE #69: FeatureInfo issues
FeatureInfo was added so that we could us the featureinfo in the before/after hooks.

This was all well and good but the examples happened to not test the before/after features and only did the before/after tests as an integration level.

The result of this was that we didn't realize that the FeatureInfo type was put under the `internal` package. This means when someone tries to actually use these beforeeachfeature hooks, the type can only be referenced as an `interface{}` since it isnt exported (since its in the internal pkg).

However, moving it into the intuitive pkg/feature location causes an import cycle.

Lastly, in trying to work with this I wanted the After hook to know about the result of the test. Did it run/pass/get skipped? If this information can be added to the FeatureInfo object that would be wonderful.

johnSchnake:
- Removed the FeatureInfo type in favor of the Feature interface
- Added logic to make a deep copy rather than pass the feature itself
- Added a test to confirm we dont allow mutation of the feature in
the before/after hooks

Fixes #69 

### DIFF
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -192,7 +192,7 @@ func (e *testEnv) Test(t *testing.T, testFeatures ...types.Feature) {
 	for _, feature := range testFeatures {
 		// execute beforeFeature actions
 		for _, action := range beforeFeatureActions {
-			if e.ctx, err = action.runWithFeature(e.ctx, e.cfg, featureInfoFromFeature(feature)); err != nil {
+			if e.ctx, err = action.runWithFeature(e.ctx, e.cfg, deepCopyFeature(feature)); err != nil {
 				t.Fatalf("BeforeEachTest failure: %s", err)
 			}
 		}
@@ -202,7 +202,7 @@ func (e *testEnv) Test(t *testing.T, testFeatures ...types.Feature) {
 
 		// execute beforeFeature actions
 		for _, action := range afterFeatureActions {
-			if e.ctx, err = action.runWithFeature(e.ctx, e.cfg, featureInfoFromFeature(feature)); err != nil {
+			if e.ctx, err = action.runWithFeature(e.ctx, e.cfg, deepCopyFeature(feature)); err != nil {
 				t.Fatalf("BeforeEachTest failure: %s", err)
 			}
 		}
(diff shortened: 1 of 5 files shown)


pkg/env/env.go
```diff
@@ -367,12 +367,16 @@ func (e *testEnv) execFeature(ctx context.Context, t *testing.T, f types.Feature
...
+// deepCopyFeature just copies the values from the Feature but creates a deep
+// copy to avoid mutation when we just want an informational copy.
+func deepCopyFeature(f types.Feature) types.Feature {
+	fcopy := features.New(f.Name())
+	for k, v := range f.Labels() {
+		fcopy = fcopy.WithLabel(k, v)
 	}
+	f.Steps()
+	for _, step := range f.Steps() {
+		fcopy = fcopy.WithStep(step.Name(), step.Level(), nil)
```
johnSchnake, line 379: Not copying the function since there isn't anything useful we need that function for. I was worried if I copied it there could be some wonkiness if it were modified/nil'd out.

pkg/env/env_test.go
```diff
@@ -398,10 +397,10 @@ func TestEnv_Test(t *testing.T) {
 			setup: func(t *testing.T, ctx context.Context) []string {
 				env := newTestEnv()
 				val := []string{}
-				env.BeforeEachFeature(func(ctx context.Context, _ *envconf.Config, info types.FeatureInfo) (context.Context, error) {
+				env.BeforeEachFeature(func(ctx context.Context, _ *envconf.Config, info features.Feature) (context.Context, error) {
 					val = append(val, "before-each-feature")
 					return ctx, nil
-				}).AfterEachFeature(func(ctx context.Context, _ *envconf.Config, info types.FeatureInfo) (context.Context, error) {
```
johnSchnake, line 404: `types` is internal; so when I tried to vendor this I could access the `AfterEachFeature` but I couldn't provide a function that referenced that `types.FeatureInfo`

pkg/env/env.go
vladimirvivien, line 379: It looks like `fcopy.WithStep` is copying the original step functions.  I think that is what should be avoided.  Maybe instead of a generic deepCopyFeature function, the function should be `deepCopyFeatureInfo` still returns a feature but does not copy the steps.

pkg/env/env_test.go
```diff
@@ -417,6 +416,49 @@ func TestEnv_Test(t *testing.T) {
...
+				"after-each-feature",
+				"before-each-feature",
+				"test-feat-2",
+				"after-each-feature",
+			},
+			setup: func(t *testing.T, ctx context.Context) []string {
+				env := newTestEnv()
+				val := []string{}
+				env.BeforeEachFeature(func(ctx context.Context, _ *envconf.Config, info features.Feature) (context.Context, error) {
+					val = append(val, "before-each-feature")
```
vladimirvivien, line 434: We just gotta make sure test developers are not able to have access and call  `info.Steps()[0](...)` step functions. If they can't, ignore this.

vladimirvivien (commented):
Looking good. Left some comments. Please take a look and make sure Feature.Steps() does not return any valid function.

pkg/env/env.go
johnSchnake: No, the last value there is the function and I'm specifically passing in `nil` instead of `step.Func()`

pkg/env/env_test.go
johnSchnake: Added an extra check or two in this test to ensure values were/werent nil as expected.

vladimirvivien (approved):
LGTM

PR #73, opened 2021-11-11
GIT-60: enable helper function for conditional waits

### LINKED ISSUE #60: Introduce support for Waiting for given cluster conditions
As a test writer of components running on Kubernetes, it would be extremely useful to have the capability to wait for one or more cluster conditions before proceeding in the test.  

### A framework for Waiting
A new package `klient/wait` could be the starting point for an API to express conditions for waiting by leveraging the `wait` package in API machinery "k8s.io/apimachinery/pkg/util/wait"):

```go
package wait

func For(cond func() (bool, error)){...}
```
So, a test writer may have condition as follows:

```go
func TestSomething(t *testing.T){
    wait.For(t, func() (bool, error) {
        var ns coreV1.Namespace
        cfg.client.List(ns)
        ...
    )}
}
```

### Pre-defined conditions
The framework could have a collection of pre-defined conditions that can be used by test writers:

```go
func TestSomething(t *testing.T) {
   var pod coreV1.Pod
   wait.For(PodReadyCondition(pod))
}

### PR DESCRIPTION
**Pull request by harshanarayana**
Closes #60

### DIFF
--- a/internal/testutil/setup.go
+++ b/internal/testutil/setup.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package testutil
+
+import (
+	"flag"
+	"k8s.io/client-go/util/homedir"
+	"log"
+	"path/filepath"
+	"time"
+
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/e2e-framework/klient/conf"
+	"sigs.k8s.io/e2e-framework/support/kind"
+)
+
+func SetupTestCluster(path string) (kc *kind.Cluster, kubeConfig string, cfg *rest.Config, clientSet kubernetes.Interface) {
+	if path == "" {
+		home := homedir.HomeDir()
+		path = filepath.Join(home, ".kube", "config")
+	}
+
+	var err error
+	kc, err = setupKind()
+	if err != nil {
+		log.Fatalln("error while setting up the kind cluster", err)
+	}
+	flag.StringVar(&kubeConfig, "kubeconfig", "", "Paths to a kubeconfig. Only required if out-of-cluster.")
+	err = flag.Set("kubeconfig", path)
+	if err != nil {
+		log.Fatalln("unexpected error while setting the flag value for kubeconfig", err)
+	}
+	flag.Parse()
+
+	cfg, err = conf.New(conf.ResolveKubeConfigFile())
+	if err != nil {
+		log.Fatalln("error while client connection trying to resolve kubeconfig", err)
+	}
+	clientSet, err = kubernetes.NewForConfig(cfg)
+	if err != nil {
+		log.Fatalln("failed to create new Client set for kind cluster", err)
+	}
+	return
+}
+
+func DestroyTestCluster(kc *kind.Cluster) {
+	err := kc.Destroy()
+	if err != nil {
+		log.Println("error while deleting the cluster", err)
+		return
+	}
+}
+
+func setupKind() (kc *kind.Cluster, err error) {
+	kc = kind.NewCluster("e2e-test-cluster")
+	if _, err = kc.Create(); err != nil {
+		return
+	}
+
+	waitPeriod := 10 * time.Second
+	log.Println("Waiting for kind pods to be initlaized...")
+	time.Sleep(waitPeriod)
+	return
+}
--- a/klient/client.go
+++ b/klient/client.go
@@ -20,6 +20,7 @@ import (
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/e2e-framework/klient/conf"
 	"sigs.k8s.io/e2e-framework/klient/k8s/resources"
+	"sigs.k8s.io/e2e-framework/klient/wait"
 )
 
 // Client stores values to interact with the
@@ -31,11 +32,16 @@ type Client interface {
 	// This method takes zero or at most 1 namespace (more will panic) that
 	// can be used in List operations.
 	Resources(...string) *resources.Resources
+	// Wait returns an implementation of the wait.Interface that can be used
+	// to perform helper operation such as waiting for certain condition to be
+	// met.
+	Wait() wait.Interface
 }
 
 type client struct {
 	cfg       *rest.Config
 	resources *resources.Resources
+	waiter    wait.Interface
 }
 
 // New returns a new Client value
@@ -44,7 +50,7 @@ func New(cfg *rest.Config) (Client, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &client{cfg: cfg, resources: res}, nil
+	return &client{cfg: cfg, resources: res, waiter: wait.New(res)}, nil
 }
 
 // NewWithKubeConfigFile creates a client using the kubeconfig filePath
@@ -73,4 +79,8 @@ func (c *client) Resources(namespace ...string) *resources.Resources {
 	default:
 		panic("too many namespaces provided")
 	}
+}
+
+func (c *client) Wait() wait.Interface {
+	return c.waiter
 }
--- a/klient/client.go
+++ b/klient/client.go
@@ -20,7 +20,6 @@ import (
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/e2e-framework/klient/conf"
 	"sigs.k8s.io/e2e-framework/klient/k8s/resources"
-	"sigs.k8s.io/e2e-framework/klient/wait"
 )
 
 // Client stores values to interact with the
@@ -32,16 +31,11 @@ type Client interface {
 	// This method takes zero or at most 1 namespace (more will panic) that
 	// can be used in List operations.
 	Resources(...string) *resources.Resources
-	// Wait returns an implementation of the wait.Interface that can be used
-	// to perform helper operation such as waiting for certain condition to be
-	// met.
-	Wait() wait.Interface
 }
 
 type client struct {
 	cfg       *rest.Config
 	resources *resources.Resources
-	waiter    wait.Interface
 }
 
 // New returns a new Client value
@@ -50,7 +44,7 @@ func New(cfg *rest.Config) (Client, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &client{cfg: cfg, resources: res, waiter: wait.New(res)}, nil
+	return &client{cfg: cfg, resources: res}, nil
 }
 
 // NewWithKubeConfigFile creates a client using the kubeconfig filePath
@@ -79,8 +73,4 @@ func (c *client) Resources(namespace ...string) *resources.Resources {
 	default:
 		panic("too many namespaces provided")
 	}
-}
-
-func (c *client) Wait() wait.Interface {
-	return c.waiter
 }
--- a/klient/internal/testutil/setup.go
+++ b/klient/internal/testutil/setup.go
@@ -17,10 +17,7 @@ limitations under the License.
 package testutil
 
 import (
-	"flag"
-	"k8s.io/client-go/util/homedir"
 	"log"
-	"path/filepath"
 	"time"
 
 	"k8s.io/client-go/kubernetes"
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

**Review by vladimirvivien — COMMENTED** [review]
@harshanarayana This is a great start and captures most of everything I was thinking about. I left some comments because I think this can be refined further.

### Design consideration
The main thing that I would like to see is to be able to invoke the `For` function from the `wait` package without the need to create a `wait.Interface instance`.  One way that could be done is to create a new subpackage `wait/cond` where all pre-conditions are defined. The `cond` package would expose `Condition` value with all the pre-conditions defined

```go
// package wait/cond
type Condition struct {
    res *resources.Resources
}
func New(*resources.Resrouces) *Condition
func (c *Condition) ResourceScaled(...)
func (c *Condition) JobMatche(...)
func (c *Condition)
...
```

Then, using the `wait` package could look like this:

```go
import ".../.../wait"
import ".../.../wait/cond"

func DoK8sStuff(){
    pod := createPod(...)
    wait.For(cond.New(*resources.Resources).PodRunning(&pod))
}
```

Hope this makes sense.

harshanarayana:
> Hope this makes sense.

Absolutely. I like the notion of adding an inlined resource so that we can use multiple difference instance of the `resources` type to our advantage as required.  Let me do the changes accordingly.

internal/testutil/setup.go
```diff
@@ -0,0 +1,79 @@
+/*
```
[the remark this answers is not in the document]
harshanarayana: Sure. Let me do that. That makes a lot of sense to keep it self contained under `klient` as there is no use of this outside of that package for now.

internal/testutil/setup.go
```diff
@@ -0,0 +1,79 @@
...
+	"path/filepath"
+	"time"
+
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/e2e-framework/klient/conf"
+	"sigs.k8s.io/e2e-framework/support/kind"
+)
+
+func SetupTestCluster(path string) (kc *kind.Cluster, kubeConfig string, cfg *rest.Config, clientSet kubernetes.Interface) {
```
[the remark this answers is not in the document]
harshanarayana: Sure. 👍

klient/client.go
```diff
@@ -31,11 +32,16 @@ type Client interface {
 	// This method takes zero or at most 1 namespace (more will panic) that
 	// can be used in List operations.
 	Resources(...string) *resources.Resources
+	// Wait returns an implementation of the wait.Interface that can be used
+	// to perform helper operation such as waiting for certain condition to be
+	// met.
+	Wait() wait.Interface
```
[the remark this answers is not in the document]
harshanarayana: The idea was, you could just do `klient.Wait().For` But if we separate the things like your initial comment suggestion and perform `wait.For` this one becomes defunct. I will take that approach and remove this method from the interface entirely.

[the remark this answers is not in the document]
harshanarayana: With `wait.For` being callable directly, there is no reason to track the `waiter` under the `klient`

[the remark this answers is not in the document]
harshanarayana: xref: ^

klient/client.go
```diff
@@ -74,3 +80,7 @@ func (c *client) Resources(namespace ...string) *resources.Resources {
 		panic("too many namespaces provided")
 	}
 }
+
+func (c *client) Wait() wait.Interface {
```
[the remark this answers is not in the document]
harshanarayana: xref: https://github.com/kubernetes-sigs/e2e-framework/pull/73#discussion_r749013514

klient/wait/wait.go
```diff
@@ -0,0 +1,139 @@
...
+	"sigs.k8s.io/e2e-framework/klient/k8s"
+	"sigs.k8s.io/e2e-framework/klient/k8s/resources"
+)
+
+const (
+	defaultPollTimeout  = 5 * time.Minute
+	defaultPollInterval = 5 * time.Second
+)
+
+type Interface interface {
```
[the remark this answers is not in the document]
harshanarayana: Sure. Let me change that accordingly. It will also help us extend this further into using `wait.PollUntilWithContext` and such in the future.

klient/wait/wait.go
```diff
@@ -0,0 +1,139 @@
...
+	return func() (done bool, err error) {
+		log.Printf("Checking for Pod %v Condition of %s/%s", phase, pod.GetNamespace(), pod.GetName())
+		if err := w.resources.Get(context.Background(), pod.GetName(), pod.GetNamespace(), pod); err != nil {
+			return false, err
+		}
+		return pod.(*v1.Pod).Status.Phase == phase, nil
+	}
+}
+
+func (w *waiter) PodReady(pod k8s.Object) apimachinerywait.ConditionFunc {
```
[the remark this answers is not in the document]
harshanarayana: Yes sir. Let me do that. I like your suggested way for this.

harshanarayana:
@vladimirvivien  I have addressed all the changes you suggested and refactored it accordingly. PTAL when you manage to get some time. Thanks

klient/internal/testutil/setup.go
```diff
@@ -0,0 +1,90 @@
...
+	KindCluster *kind.Cluster
+	Kubeconfig string
+	RESTConfig *rest.Config
+	Clientset kubernetes.Interface
+}
+
+func SetupTestCluster(path string) *TestCluster {
+	if path == "" {
+		home := homedir.HomeDir()
+		path = filepath.Join(home, ".kube", "config")
```
vladimirvivien, line 42: You may be interested in using `conf.ResolveKubeConfigFile` (found here https://github.com/kubernetes-sigs/e2e-framework/blob/main/klient/conf/config.go)

klient/wait/conditions/conditions.go
```diff
@@ -0,0 +1,119 @@
...
+	"log"
+	"sigs.k8s.io/e2e-framework/klient/k8s"
+	"sigs.k8s.io/e2e-framework/klient/k8s/resources"
+)
+
+type Condition struct {
+	resources *resources.Resources
+}
+
+func New(resources *resources.Resources) *Condition {
```
vladimirvivien, line 34: Please add documentation comments for functions.

klient/wait/conditions/conditions.go
```diff
@@ -0,0 +1,119 @@
...
+	return &Condition{resources: resources}
+}
+
+func (c *Condition) ResourceScaled(obj k8s.Object, scaleFetcher func(object k8s.Object) int32, replica int32) apimachinerywait.ConditionFunc {
```
vladimirvivien, line 38: Add doc comment for methods as well.

klient/wait/conditions/conditions.go
```diff
@@ -0,0 +1,119 @@
...
+	return func() (done bool, err error) {
+		log.Printf("Checking if the resource %s/%s has been scaled to %d", obj.GetNamespace(), obj.GetName(), replica)
```
vladimirvivien, line 40: These log statements will be a bit verbose when used.  I would consider removing them.

klient/wait/main_test.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func tearDown() {
+	deleteNamespace()
+	tc.DestroyTestCluster()
+}
+
+func deleteNamespace() {
+	namespace := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}}
+	_ = getResourceManager().Delete(context.TODO(), namespace)
```
vladimirvivien, line 74: I understand why error is not handled (because it's used during tear down), but I would suggest return the error here and ignore it when it's used (in `tearDown`)

klient/wait/wait_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"sigs.k8s.io/e2e-framework/klient/k8s"
+	"sigs.k8s.io/e2e-framework/klient/wait/conditions"
+	"testing"
+	"time"
+)
+
+func TestPodRunning(t *testing.T) {
+	var err error
+	pod := createPod("p1", t)
+	err = For(conditions.New(getResourceManager()).PodRunning(pod), WithImmediate())
```
vladimirvivien, line 33: This is great!!! That's what I was looking for.

klient/wait/conditions/conditions.go
harshanarayana: Can I make these optional? By default it will be turned off but can be enabled by the users if required. Would that be ok ?

klient/wait/conditions/conditions.go
harshanarayana: Updating the Documentation for all of them. Thanks for the heads up

klient/wait/conditions/conditions.go
harshanarayana: Updating the Documentation for all of them. Thanks for the heads up

klient/internal/testutil/setup.go
harshanarayana: Yes. I am using that already and like you mentioned below, I think we can remove Line 52-57 and this one too.

klient/internal/testutil/setup.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	tc := &TestCluster{}
+	var err error
+	kc, err := setupKind()
+	if err != nil {
+		log.Fatalln("error while setting up the kind cluster", err)
+	}
+	tc.KindCluster = kc
+	flag.StringVar(&tc.Kubeconfig, "kubeconfig", "", "Paths to a kubeconfig. Only required if out-of-cluster.")
```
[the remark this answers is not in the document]
harshanarayana: You are right. We can get away with not having this. Let me clean this up.

klient/wait/main_test.go
harshanarayana: Let me update this.

harshanarayana:
> First apologies for the late review.

No worries at all @vladimirvivien . Thanks again for taking your time to review this. I have addressed the changes you requested. Let me know if you think anything else needs to be modified

klient/wait/conditions/conditions.go
vladimirvivien: I think outputting logs from the framework code could be an issue because, as a user, when I write tests, I don't expect extra output unless I use something like t.Log (etc).

As far as making it optional, here is what  we can do:
* Get this merged (without logs)
* Create a new PR to introduce more robust log capabilities via a level log framework 
* Then we can research for something that make sense for the project.
How does that sound ?

klient/wait/wait_test.go
```diff
@@ -0,0 +1,116 @@
...
+	v1 "k8s.io/api/core/v1"
+
+	"sigs.k8s.io/e2e-framework/klient/k8s"
+	"sigs.k8s.io/e2e-framework/klient/wait/conditions"
+)
+
+func TestPodRunning(t *testing.T) {
+	var err error
+	pod := createPod("p1", t)
+	err = For(conditions.New(getResourceManager()).WithVerboseLog().PodRunning(pod), WithImmediate())
```
vladimirvivien, line 35: I don't think it's a good idea to control logging this way.
Log control should be independent of that the `wait` package
and should be done universally the same way.

klient/wait/conditions/conditions.go
```diff
@@ -0,0 +1,164 @@
...
+
+// New is used to create a new Condition that can be used to perform a series of pre-defined wait checks
+// against a resource in question
+func New(r *resources.Resources) *Condition {
+	return &Condition{resources: r}
+}
+
+// WithVerboseLog is used to enable additional verbose logs from the Wait condition to help debug tests
+func (c *Condition) WithVerboseLog() *Condition {
+	c.verbose = true
```
vladimirvivien, line 48: I understand your concern about ability to see messages from long running tests.
I just don't think log control should be done this way using the wait API.
Log should be controlled universally during test runtime and should be handled in a separate PR.

Suggestion:
* Put back the hard coded `log.Println` statement as you had before
* We'll get it merged
* Then I can open an issue to revisit logging universally for the entire project/package

vladimirvivien (commented):
@harshanarayana 
I still have reservation about how logging is done (it feels a bit one-off). I left comments.
Other than than that, we should be ready to merge soon.

klient/wait/conditions/conditions.go
harshanarayana: If it is all right with you, can we do it the other way? I will cleanup the logs for now and have nothing in it., We can revisit this again with the new Issue and tackle that as a whole thing generally.. 
I will update the PR to cleanup the logging for now.

klient/wait/wait_test.go
harshanarayana: xref: https://github.com/kubernetes-sigs/e2e-framework/pull/73#discussion_r752862420

klient/wait/conditions/conditions.go
harshanarayana: Opened https://github.com/kubernetes-sigs/e2e-framework/issues/77 to track the same.

vladimirvivien (approved):
@harshanarayana Awesome work.

vladimirvivien:
@harshanarayana 
Couple things I just realized we need
* Add an example test in the examples dir
* A quick documentation blurb in README.md
Please open a new issue you or someone else can get that done.

Thanks again!!!

harshanarayana:
@vladimirvivien let me take care of it right away.

harshanarayana:
> * Please open a new issue you or someone else can get that done.

https://github.com/kubernetes-sigs/e2e-framework/issues/78

helayoty:
This is a very needed feature, thanks @harshanarayana for the great effort. @vladimirvivien When do we expect a new release so we can use it?

Issue #78, opened 2021-11-19
Add Documentation for wait/Conditions


harshanarayana:
Opening this item to track the comment suggested by @vladimirvivien  under https://github.com/kubernetes-sigs/e2e-framework/pull/73#issuecomment-974213738

The `sigs.k8s.io/e2e-framework/klient/wait` needs an example under the [examples](https://github.com/kubernetes-sigs/e2e-framework/tree/main/examples) folder and a `README` added to explain what the tests are doing.

harshanarayana:
Changes were addressed in #79

PR #79, opened 2021-11-22
add condition helpers and examples

crandles:
While reviewing the conditions package I noticed there weren't any functions to help deal with lists of objects nor helpers built around `k8s.Object` to deal with types outside of pod, jobs, and other core types.

I think I should be able to use any type in my codebase now.

Saw #78 --  so I tried to add some examples as well.

### DIFF
--- a/klient/wait/conditions/conditions.go
+++ b/klient/wait/conditions/conditions.go
@@ -22,9 +22,11 @@ import (
 
 	log "k8s.io/klog/v2"
 
+	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	apimachinerywait "k8s.io/apimachinery/pkg/util/wait"
 
 	"sigs.k8s.io/e2e-framework/klient/k8s"
@@ -55,6 +57,124 @@ func (c *Condition) ResourceScaled(obj k8s.Object, scaleFetcher func(object k8s.
 			return false, nil
 		}
 		return scaleFetcher(obj) == replica, nil
+	}
+}
+
+// ResourceMatch is a helper function used to check if the resource under question has met a pre-defined state. This can
+// be leveraged for checking fields on a resource that may not be immediately present upon creation.
+func (c *Condition) ResourceMatch(obj k8s.Object, matchFetcher func(object k8s.Object) bool) apimachinerywait.ConditionFunc {
+	return func() (done bool, err error) {
+		if err := c.resources.Get(context.TODO(), obj.GetName(), obj.GetNamespace(), obj); err != nil {
+			return false, nil
+		}
+		return matchFetcher(obj), nil
+	}
+}
+
+// ResourceListN is a helper function that can be used to check for a minimum number of returned objects in a list. This function
+// accepts list options that can be used to adjust the set of objects queried for in the List resource operation.
+func (c *Condition) ResourceListN(list k8s.ObjectList, n int, listOptions ...resources.ListOption) apimachinerywait.ConditionFunc {
+	return c.ResourceListMatchN(list, n, func(object k8s.Object) bool { return true }, listOptions...)
+}
+
+// ResourceListMatchN is a helper function that can be used to check for a minimum number of returned objects in a list. This function
+// accepts list options and a match function that can be used to adjust the set of objects queried for in the List resource operation.
+func (c *Condition) ResourceListMatchN(list k8s.ObjectList, n int, matchFetcher func(object k8s.Object) bool, listOptions ...resources.ListOption) apimachinerywait.ConditionFunc {
+	return func() (done bool, err error) {
+		if err := c.resources.List(context.TODO(), list, listOptions...); err != nil {
+			return false, nil
+		}
+		var found int
+		metaList, err := meta.ExtractList(list)
+		if err != nil {
+			return false, err
+		}
+		for _, obj := range metaList {
+			if o, ok := obj.(k8s.Object); ok && matchFetcher(o) {
+				found++
+			} else if !ok {
+				return false, fmt.Errorf("condition: unexpected type %T in list, does not satisfy k8s.Object", obj)
+			}
+		}
+		return found >= n, nil
+	}
+}
+
+// ResourcesFound is a helper function that can be used to check for a set of objects. This function accepts a list
+// of named objects and will wait until it is able to retrieve each.
+func (c *Condition) ResourcesFound(list k8s.ObjectList) apimachinerywait.ConditionFunc {
+	return c.ResourcesMatch(list, func(object k8s.Object) bool { return true })
+}
+
+// ResourcesMatch is a helper function that can be used to check for a set of objects. This function accepts a list
+// of named objects and a match function, and will wait until it is able to retrieve each while passing the match validation.
+func (c *Condition) ResourcesMatch(list k8s.ObjectList, matchFetcher func(object k8s.Object) bool) apimachinerywait.ConditionFunc {
+	metaList, err := meta.ExtractList(list)
+	if err != nil {
+		return func() (done bool, err error) { return false, err }
+	}
+	objects := make(map[k8s.Object]bool)
+	for _, o := range metaList {
+		obj, ok := o.(k8s.Object)
+		if !ok {
+			return func() (done bool, err error) {
+				return false, fmt.Errorf("condition: unexpected type %T in list, does not satisfy k8s.Object", obj)
+			}
+		}
+		if obj.GetName() != "" {
+			objects[obj] = false
+		}
+	}
+	return func() (done bool, err error) {
+		found := 0
+		for obj, created := range objects {
+			if !created {
+				if err := c.resources.Get(context.TODO(), obj.GetName(), obj.GetNamespace(), obj); errors.IsNotFound(err) {
+					continue
+				} else if err != nil {
+					return false, err
+				}
+				if !matchFetcher(obj) {
+					continue
+				}
+			}
+			objects[obj] = true
+			found++
+		}
+		return len(objects) == found, nil
+	}
+}
+
+// ResourcesDeleted is a helper function that can be used to check for if a set of objects has been deleted. This function
+// accepts a list of named objects and will wait until it is not able to find each.
+func (c *Condition) ResourcesDeleted(list k8s.ObjectList) apimachinerywait.ConditionFunc {
+	metaList, err := meta.ExtractList(list)
+	if err != nil {
+		return func() (done bool, err error) { return false, err }
+	}
+	objects := make(map[k8s.Object]bool)
+	for _, o := range metaList {
+		obj, ok := o.(k8s.Object)
+		if !ok {
+			return func() (done bool, err error) {
+				return false, fmt.Errorf("condition: unexpected type %T in list, does not satisfy k8s.Object", obj)
+			}
+		}
+		if obj.GetName() != "" {
+			objects[obj] = true
+		}
+	}
+	return func() (done bool, err error) {
+		for obj, created := range objects {
+			if created {
+				if err := c.resources.Get(context.TODO(), obj.GetName(), obj.GetNamespace(), obj); errors.IsNotFound(err) {
+					delete(objects, obj)
+				} else if err != nil {
+					return false, err
+				}
+			}
+		}
+		return len(objects) == 0, nil
 	}
 }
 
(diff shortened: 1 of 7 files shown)


vladimirvivien (commented):
@crandles This is awesome work.
This contribution will be super useful to the community.
I left a few comments (mostly naming of things).

### A generic Matcher
While reviewing your code, i realize that there might be a need for a generic ConditionMatch method that would allow users to specify any condition check on a given resource. However, as I dug into it, I realize you need a concrete resource type to get to the resource status/condition:

```go
func (c *Condition)  ConditionMatch(obj k8s.Object, fn func(interface{})(bool, error)) apimachinerywait.ConditionFunc {
	return func() (done bool, err error) {
		if err := c.resources.Get(context.TODO(), obj.GetName(), obj.GetNamespace(), obj); err != nil {
			return false, err
		}
                // need concrete type to get to stats/condition
		return
	}
}
```

Leaving the above, maybe you have a better idea. If not, no worries.

crandles:
> ### A generic Matcher
> While reviewing your code, i realize that there might be a need for a generic ConditionMatch method that would allow users to specify any condition check on a given resource. However, as I dug into it, I realize you need a concrete resource type to get to the resource status/condition:
> 
> ```go
> func (c *Condition)  ConditionMatch(obj k8s.Object, fn func(interface{})(bool, error)) apimachinerywait.ConditionFunc {
> 	return func() (done bool, err error) {
> 		if err := c.resources.Get(context.TODO(), obj.GetName(), obj.GetNamespace(), obj); err != nil {
> 			return false, err
> 		}
>                 // need concrete type to get to stats/condition
> 		return
> 	}
> }
> ```
> 
> Leaving the above, maybe you have a better idea. If not, no worries.

This looks pretty close to this -- you do need to do a type cast, but you can push that to the caller to define: 

https://github.com/kubernetes-sigs/e2e-framework/blob/ecc937d52b9890c32e7bba475dae45ccd18a9446/klient/wait/conditions/conditions.go#L55-L64

https://github.com/kubernetes-sigs/e2e-framework/blob/ecc937d52b9890c32e7bba475dae45ccd18a9446/klient/wait/wait_test.go#L223-L234

klient/wait/conditions/conditions.go
```diff
@@ -50,6 +52,112 @@ func (c *Condition) ResourceScaled(obj k8s.Object, scaleFetcher func(object k8s.
...
+		if err := c.resources.Get(context.TODO(), obj.GetName(), obj.GetNamespace(), obj); err != nil {
+			return false, nil
+		}
+		return matchFetcher(obj), nil
+	}
+}
+
+// ListResources is a helper function that can be used to check for a minimum number of returned objects in a list. This function
+// accepts list options that can be used to adjust the set of objects queried for in the List resource operation.
+func (c *Condition) ListResources(list k8s.ObjectList, n int, listOptions ...resources.ListOption) apimachinerywait.ConditionFunc {
```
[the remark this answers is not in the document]
crandles: Sounds great, I was hoping for some input on these names.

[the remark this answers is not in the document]
crandles: Renamed

harshanarayana:
This is brilliant set of changes. Thanks for making it generic.

vladimirvivien (approved):
LGTM!
Thanks a lot for a great contribution.

crandles:
I do not think I modified any code used by this failing test:
```
ok  	sigs.k8s.io/e2e-framework/klient/k8s/resources	42.400s	coverage: 56.6% of statements
I1202 16:34:05.472403   29103 setup.go:78] Waiting for kind pods to be initialized...
=== RUN   TestPodRunning
    main_test.go:102: failed to create pod due to an error pods "p1" is forbidden: error looking up service account wait-test/default: serviceaccount "default" not found
    wait_test.go:41: failed to wait for pod to reach running condition pods "p1" not found
--- FAIL: TestPodRunning (1.47s)
=== RUN   TestPodPhaseMatch
--- PASS: TestPodPhaseMatch (25.74s)
=== RUN   TestPodReady
--- PASS: TestPodReady (2.01s)
=== RUN   TestContainersReady
skipped 26 lines unfold_more
I1202 16:36:18.889256   29103 wait_test.go:221] Done
--- PASS: TestResourcesDeleted (20.03s)
=== RUN   TestResourceMatch
I1202 16:36:23.899623   29103 wait_test.go:234] Done
--- PASS: TestResourceMatch (5.01s)
FAIL
coverage: 94.4% of statements
FAIL	sigs.k8s.io/e2e-framework/klient/wait	169.541s
```

But will investigate. Maybe there is a flake?

harshanarayana:
seems like a flaky test failure. 
xref: https://github.com/kubernetes/kubernetes/issues/66689

crandles:
> seems like a flaky test failure. xref: [kubernetes/kubernetes#66689](https://github.com/kubernetes/kubernetes/issues/66689)

That is interesting -- with this PR I could actually fix this problem by adding a "wait for service account to be available"

crandles:
I noticed that when kind starts up some components have not completed started that are important (networking, kube-system controllers, etc).

harshanarayana:
> I noticed that when kind starts up some components have not completed starting that are important (networking, kube-system controllers, etc).
> 
> You may be able to create resources with the API but they may not be processed until key components start.
> 
> * if the controller-manager is down, replicasets and pods will not be generated from Deployments
> * If the networking is down, your pods might not be running
> * If you are installing CRDs and operators, this compounds (if you haven't waited for the above to start)
> * etc

```go
func setupKind() (kc *kind.Cluster, err error) {
	kc = kind.NewCluster("e2e-test-cluster")
	if _, err = kc.Create(); err != nil {
		return
	}

	waitPeriod := 10 * time.Second
	log.Info("Waiting for kind pods to be initialized...")
	time.Sleep(waitPeriod)
	return
}
```
this one does a plain wait after the cluster is created. Might be a good idea to enhance this to check for atleast control plane to be fully up before we return off this method ? API Server + Controller Manager + Scheduler at minimum might be sufficient ?

harshanarayana:
Thought about using the `--wait` arg of the `kind` but that has a similar effect. there is no gurantee the timeout is good enough. Better to poll and check if things are up before yielding the control back ?

crandles:
Added a simple "wait for control-plane core components" to _two_ instances (internal and envfuncs) of cluster creation used during tests.

vladimirvivien:
Very happy to see this thread!!
Thanks for the fix.

pkg/envfuncs/kind_funcs.go
```diff
@@ -68,16 +76,41 @@ func CreateKindClusterWithConfig(clusterName, image, configFilePath string) env.
...
+		if err := waitForControlPlane(ctx, cfg.Client()); err != nil {
+			return ctx, err
+		}
+
 		// store entire cluster value in ctx for future access using the cluster name
 		return context.WithValue(ctx, kindContextKey(clusterName), k), nil
 	}
 }
 
+func waitForControlPlane(ctx context.Context, client klient.Client) error {
```
vladimirvivien, line 92: This is awesome !!

vladimirvivien: Suggestion: maybe create a pre-built condition wrapper method 

```go
func (c *Condition) ControlPlaneReady(...)
```
Just a suggestion, not necessary, but since the logic is reusable, it may be a nice to have.

vladimirvivien (approved):
This is awesome work.
I left couple of comments.
Ping me when done and I will approve.

crandles:
Running into the same error with the wait-test/default service account not being available.. not yet sure how to fix / it still does not seem related to code I've changed.

There are a few more pods I can try to wait for (kind networking pods); will see if that helps. Perhaps the 7 second wait I removed helped somehow (?) I am missing

crandles:
@vladimirvivien this could use another test to see if the additional wait helps.

pkg/envfuncs/kind_funcs.go
crandles: Right now I have the control-plane readiness defined twice; once in an internal test for the klient/wait package, and another in the envfuncs package.

Creating a re-usable func on the condition type might allow me to remove the duplication -- will think on that. I had avoided it at first out of fear of an import cycle but will revisit.

vladimirvivien:
@crandles 
The test flakiness may have to do with the test infrastructure CI/CD where this is is executed (not the code it self).
Please squash commits to get this approved.

crandles:
@vladimirvivien done 🚢

klient/wait/conditions/conditions.go
```diff
@@ -58,6 +60,120 @@ func (c *Condition) ResourceScaled(obj k8s.Object, scaleFetcher func(object k8s.
...
+func (c *Condition) ResourcesDeleted(list k8s.ObjectList) apimachinerywait.ConditionFunc {
+	metaList, err := meta.ExtractList(list)
+	if err != nil {
+		return func() (done bool, err error) { return false, err }
+	}
+	objects := make(map[k8s.Object]bool)
+	for _, o := range metaList {
+		obj, ok := o.(k8s.Object)
+		if !ok {
+			panic(fmt.Errorf("unexpected type %T in list, does not satisfy k8s.Object", obj))
```
vladimirvivien, line 157: I don't think this should panic, just an error return should be ok.

klient/wait/conditions/conditions.go
```diff
@@ -58,6 +60,120 @@ func (c *Condition) ResourceScaled(obj k8s.Object, scaleFetcher func(object k8s.
 	}
 }
 
+// ResourceMatch is a helper function used to check if the resource under question has met a pre-defined state. This can
+// be leveraged for checking fields on a resource that may not be immediately present upon creation.
+func (c *Condition) ResourceMatch(obj k8s.Object, matchFetcher func(object k8s.Object) bool) apimachinerywait.ConditionFunc {
+	return func() (done bool, err error) {
+		if err := c.resources.Get(context.TODO(), obj.GetName(), obj.GetNamespace(), obj); err != nil {
```
vladimirvivien, line 67: 😢  My apologies for not catching this in earlier reviews.
The `Condition` methods, that act on the resources, should receive context (similar to how methods on `Resources` work):

```go
func (c *Condition) ResourceMatch(ctx context.Context, obj k8s.Object, matchFetcher func(object k8s.Object) bool) apimachinerywait.ConditionFunc {...}
```

That way, when a resource method is invoked, the resource can be passed in (instead of creating a `context.TODO()`):

```go
if err := c.resources.Get(ctx, obj.GetName(), obj.GetNamespace(), obj); err != nil {...}
```

klient/wait/conditions/conditions.go
```diff
@@ -58,6 +60,120 @@ func (c *Condition) ResourceScaled(obj k8s.Object, scaleFetcher func(object k8s.
...
+		var found int
+		metaList, err := meta.ExtractList(list)
+		if err != nil {
+			return false, err
+		}
+		for _, obj := range metaList {
+			if o, ok := obj.(k8s.Object); ok && matchFetcher(o) {
+				found++
+			} else if !ok {
+				return false, fmt.Errorf("unexpected type %T in list, does not satisfy k8s.Object", obj)
```
vladimirvivien, line 96: Suggestion: add a context to this error (and other errors like it), something like:

```go
fmt.Errorf("condition: resource match: type %T does not satisfy k8s.Object")
```

vladimirvivien (commented):
😭 We were so close on getting this one in. Unfortunately I found couple issues.  One of them is my bad for not catching earlier (passing in context).  The others are more esthetic.

klient/wait/conditions/conditions.go
crandles: That would throw off the usage of condition slightly, if it returned (apimachinerywait.ConditionFunc, error) compared to other functions.

Or perhaps I could emit the error in the `ConditionFunc`?

(will address this in ResourcesMatch too)

klient/wait/conditions/conditions.go
crandles: I am leaning towards returning a ConditionFunc that simply exits with the appropriate error

klient/wait/conditions/conditions.go
crandles: I am leaning towards returning a ConditionFunc that simply exits with the appropriate error

klient/wait/conditions/conditions.go
crandles: Looking at this a little further, before I diverge the function signature from the rest of the package.

It seems like we could pass in `WithStopChannel(ctx.Done())` to `wait.For` and solve this, instead of having to add to the condition function here? 

```go
wait.For(conditions.New(resources).ResourceDeleted(...), WithStopChannel(ctx.Done()))
```

klient/wait/conditions/conditions.go
vladimirvivien: @crandles Yes, that's what I meant, return ConditionFunc with an error, you are already doing it in other parts of the code.

klient/wait/conditions/conditions.go
vladimirvivien: @crandles I would rather pass in the context when the condition method is invoked instead of capturing the context too early. That way, it allows users of the method to adjust the context at any point (where as if it's captured early, then it cannot be modified).

Suggestion:
For sake of getting this PR in (i'd like a release soon with all the goodness you contributed), let's make that signature correction in a different PR. 

What do you think ?

crandles: That makes sense to me -- we can adjust all of the signatures to include context at that point.

klient/wait/wait_test.go
```diff
@@ -115,3 +119,121 @@ func TestResourceScaled(t *testing.T) {
...
+	log.Info("Done")
+}
+
+func TestResourceMatch(t *testing.T) {
+	var err error
+	deployment := createDeployment("d6", 2, t)
+	err = For(conditions.New(getResourceManager()).ResourceMatch(deployment, func(object k8s.Object) bool {
+		d, ok := object.(*appsv1.Deployment)
+		if !ok {
+			panic(fmt.Errorf("unexpected type %T in list, does not satisfy *appsv1.Deployment", object))
```
vladimirvivien, line 231: Suggestion: instead of panic, use `t.Fatal`

vladimirvivien (commented):
Added one more suggestion. Everything else is good and this should merge today.

vladimirvivien (approved):
Awesome job @crandles

harshanarayana:
I am so glad to see this one in :) #LetsPlay

PR #84, opened 2021-11-30
use stdout from kind command when writing kubeconfig

crandles:
When running kind with podman, an extra log appears that is getting picked up by the `support/kind` package and breaking:

Example:
```bash
# kind version
kind v0.11.1 go1.16.4 linux/amd64
# kind get clusters
enabling experimental podman provider
myclustername1
```

https://github.com/kubernetes-sigs/kind/blob/fa7d86470f4c0e924fc4c2e767ec8491c45f4304/pkg/cluster/internal/providers/podman/provider.go#L45

This is output when running `kind get kubeconfig` to stderr -- the current processes combine stdout + stderr and results in an invalid yaml file.

This change writes only the stdout contents to the kubeconfig file. An alternative might be to run `kind get kubeconfig` with the `-q` flag.

### DIFF
--- a/support/kind/kind.go
+++ b/support/kind/kind.go
@@ -17,6 +17,7 @@ limitations under the License.
 package kind
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


vladimirvivien (approved):
Thank you!!

PR #88, opened 2021-12-02
GIT-49: enable parallel run of test features

### LINKED ISSUE #49: Execute test features in parallel
Currently all tests are executed serially. This issue is to request that the framework also support parallel test execution.

> This originated from #17 

### Feature testing
A feature is considered to be a unit of testable logic in the code. As such, the step functions, that make up a feature, should always be executed in serial as a unit.  This guarantees predictable execution flow and predictable context propagation.

```go
TestFunc(t *testing.T){
    f := feature.New("my feature").Assess("check1", ...).Assess("check2"...).Feature()
    env.Test(f)
}
```
In the previous snippet, assessments "check1" and "check2" will be executed serially as part of the feature.

### Concurrent feature testing
A test function with multiple features, however, should be able to be exercised concurrently as shown below:

```go
    f0 := feature.New("my feature").Assess("check1", ...).Assess("check2"...).Feature()
    f1 := feature.New("my feature").Assess("check3", ...).Assess("check4"...).Feature()
    f2 := feature.New("my feature").Assess("check5", ...).Assess("check6"...).Feature()

    env.TestInParallel(f0, f1, f2)
```
Features f0, f1, f2 should be executed concurrently.

> Note `env.Test(f)` and `env.TestInParallel(f)` should be equivalent.

### Forcing concurrent execution
It is convenient to be able to force tests to be executed concurrently.  The code should support the ability to execute all tests in a package concurrently using an environment configuration.

This can be done programmatically as shown below:
```go
TestMain(m *testing.M){
    env  := env.NewWithConfig(envconfig.New().WithParallelTestEnabled())
}
```

Or driven by the `--parallel` flag, by creating the environment configuration from CLI flags as shown:

```go
TestMain(m *testing.M){
    env  := env.NewWithConfig(envconfig.NewFromFlags())
}
```
Then, tests are executed as

```
go test ./package -args --parallel
```

> When the configured for parallel testing `env.Test(f0, f1, f2)` and `env.TestInParallel(f0, f1, f2)` are equivalent

harshanarayana:
closes #49 

### TODO

- [x] Add Unit Tests
- [x] Add Documentation
- [x] Add Examples

### DIFF
--- a/examples/parallel_features/README.md
+++ b/examples/parallel_features/README.md
@@ -1,0 +1,19 @@
+# Parallel Run of Test Features
+
+This directory contains the example of how to run the features in parallel as part of your test run using
+the right flags to the test runtime
+
+# Running tests with flags
+
+These tests can be executed using the normal `go test` command by passing the right arguments
+
+```bash
+go test -v . -args --parallel
+```
+
+One can also build a test binary and run that with the right arguments.
+
+```bash
+go test -c -o parallel.test .
+./parallel.test --parallel
+```
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"sync"
 	"testing"
 	"time"
 
@@ -50,6 +51,10 @@ type testEnv struct {
 // New creates a test environment with no config attached.
 func New() types.Environment {
 	return newTestEnv()
+}
+
+func NewParallel() types.Environment {
+	return newTestEnvWithParallel()
 }
 
 // NewWithConfig creates an environment using an Environment Configuration value
@@ -93,6 +98,13 @@ func newTestEnv() *testEnv {
 		ctx: context.Background(),
 		cfg: envconf.New(),
 		rnd: rand.NewSource(time.Now().UnixNano()),
+	}
+}
+
+func newTestEnvWithParallel() *testEnv {
+	return &testEnv{
+		ctx: context.Background(),
+		cfg: envconf.New().WithParallelTestEnabled(),
 	}
 }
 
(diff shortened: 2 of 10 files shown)


harshanarayana:
@crandles @vladimirvivien Please review this change when time permits and let me know if this is in the right direction.

crandles (commented):
I think this looks ok but still comparing this against parallel ginkgo / stdlib tests

examples/parallel_features/README.md
```diff
@@ -0,0 +1,19 @@
...
+These tests can be executed using the normal `go test` command by passing the right arguments
+
+```bash
+go test -v . -args --parallel
+```
+
+One can also build a test binary and run that with the right arguments.
+
+```bash
+go test -c -o paralles.test .
```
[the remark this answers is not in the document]
harshanarayana: Corrected. Thanks @crandles

harshanarayana:
> still comparing this against parallel ginkgo / stdlib tests

I will have to dig into a bit more detail of how `ginkgo` does parallel tests runs inside a given test suite. Been a while since I was in that part of the woods. 

The std benchmarking test however follows a similar approach to what I have here for running parallel tests

```go
for p := 0; p < numProcs; p++ {
		go func() {
			defer wg.Done()
			pb := &PB{
				globalN: &n,
				grain:   grain,
				bN:      uint64(b.N),
			}
			body(pb)
		}()
	}
```
I thought about being able to reuse the `t.Parallel()` helper to get the behavior I wanted, but that did not work as I intended for it to. I am open to suggestion on what else might be a better way to get to achieve this.

pkg/env/env.go
```diff
@@ -160,70 +172,134 @@ func (e *testEnv) AfterEachTest(funcs ...types.TestEnvFunc) types.Environment {
...
 		}
 	}
+}
+
+// processTestFeature is used to trigger the execution of the actual feature. This function wraps the entire
+// workflow of orchestrating the feature execution be running the action configured by BeforeEachFeature /
+// AfterEachFeature. This can be invoked with a nillable wg which defines how the test feature is being run.
+// If the wg is set to nil that means the test is being run in sequence and there is no reason to mark a
+// WaitGroup as done
+func (e *testEnv) processTestFeature(wg *sync.WaitGroup, t *testing.T, featureName string, feature types.Feature) {
```
vladimirvivien, line 199: I don't think this should take a workgroup. I suggest to keep generic as possible for clarity (see how in next comment).

pkg/env/env.go
```diff
@@ -160,70 +172,134 @@ func (e *testEnv) AfterEachTest(funcs ...types.TestEnvFunc) types.Environment {
...
-			t.Fatalf("AfterEachTest failure: %s", err)
+
+	e.processTestActions(t, beforeTestActions)
+
+	if e.cfg.ParallelTestEnabled() && enableParallelRun {
+		log.V(4).Info("Running test features in parallel")
+	}
+
+	var wg sync.WaitGroup
+	for i, feature := range testFeatures {
```
vladimirvivien, line 249: Suggestion: I would separate the execution paths (parallel, non-parallel) cleanly for clarity.  Also, this would make method `e.processTestFeature` truly generic without the need to pass a nil.

```go
if parallelEnabled {
    var wg sync.WaitGroup
    for i, feature := range testFeatures {
        wg.Add(1)
        go func(w *sync.Workgroup){
            defer w.Done()
            e.processTestFeature(t, featName, feature)
          }(&wg)
    }
    wg.Wait()
}else{
    for i, feature := range testFeatures {
         e.processTestFeature(t, featName, feature)
    }
}
```

pkg/env/env.go
```diff
@@ -160,70 +172,134 @@ func (e *testEnv) AfterEachTest(funcs ...types.TestEnvFunc) types.Environment {
...
+			e.processTestFeature(nil, t, featName, feature)
 		}
 	}
+	if e.cfg.ParallelTestEnabled() && enableParallelRun {
+		wg.Wait()
+	}
+	e.processTestActions(t, afterTestActions)
+}
+
+// TestInParallel executes a series a feature tests from within a
```
vladimirvivien, line 267: ❤️ the comments on these methods.

vladimirvivien (commented):
This is an awesome start and thank you for the effort.
I left some comments.

pkg/env/env.go
```diff
@@ -160,70 +172,134 @@ func (e *testEnv) AfterEachTest(funcs ...types.TestEnvFunc) types.Environment {
...
+	}
+	beforeTestActions := e.getBeforeTestActions()
 	afterTestActions := e.getAfterTestActions()
-	for _, action := range afterTestActions {
-		if e.ctx, err = action.runWithT(e.ctx, e.cfg, t); err != nil {
-			t.Fatalf("AfterEachTest failure: %s", err)
+
+	e.processTestActions(t, beforeTestActions)
+
+	if e.cfg.ParallelTestEnabled() && enableParallelRun {
```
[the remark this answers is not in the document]
harshanarayana: Sure. I will take care of it.

pkg/env/env.go
harshanarayana: You have a valid point. The `porcessTestFeature` can/should indeed be agnostic of the `WaitGroup` and the sequence/parallel workflow. What you suggested here looks better. Let me update it accordingly.

harshanarayana:
```golang
for i, feature := range testFeatures {
		featName := feature.Name()
		if featName == "" {
			featName = fmt.Sprintf("Feature-%d", i+1)
		}
		if runInParallel {
			wg.Add(1)
			go func(wg *sync.WaitGroup) {
				defer wg.Done()
				e.processTestFeature(t, featName, feature)
			}(&wg)
		} else {
			e.processTestFeature(t, featName, feature)
		}
	}
	if runInParallel {
		wg.Wait()
	}
```

turned it like above to avoid duplication of loop + the feature name generation bit.

harshanarayana:
> This is an awesome start and thank you for the effort. I left some comments.

I have addressed the review comments you have. Please check then when time permits. Let me know if you see any other changes required as well

pkg/env/env.go
```diff
@@ -160,70 +172,133 @@ func (e *testEnv) AfterEachTest(funcs ...types.TestEnvFunc) types.Environment {
...
-// are executed as a subtests of the feature.  This approach allows
-// features/assessments to be filtered using go test -run flag.
-//
-// Feature tests will have access to and able to update the context
-// passed to it.
-//
-// BeforeTest and AfterTest operations are executed before and after
-// the feature is tested respectively.
-func (e *testEnv) Test(t *testing.T, testFeatures ...types.Feature) {
+// panicOnMissingContext is used to check if the test Env has the right context setup
```
vladimirvivien, line 175: Nit: it checks not for the `right` context, but that there is a `non-nil` context.

vladimirvivien (commented):
@harshanarayana This is great work and progress on this PR.
Couple of things:
* Add test in pkg/envconf/config_test.go
* Add test in pkg/flags/flags_test.go

Also left comment.

harshanarayana:
@vladimirvivien  Happy new year. 

I just got back to working mode and will take a look at this today and get your comments addressed.

harshanarayana:
@vladimirvivien  I've updated the PR and added tests for the two files that were missing earlier in the PR.

vladimirvivien (approved):
Thanks @harshanarayana . This looks awesome.

PR #90, opened 2021-12-06
GIT-83: enable helm support for test workflow

### LINKED ISSUE #83: Support Helm chart
As a test writer, I should be able to deploy `helm` charts.

# Design

The framework should provide a type that allows test writers to install `helm charts`

```
type Helm []struct {
	Name                      string
	Namespace            string
        ReleaseName         string
        Version                    string
}
```

Type Helm should also include a method to install the helm chart

# Example

```
func TestMain(m *testing.M){
	testenv = env.New()
	kindClusterName := envconf.RandomName("kind-with-config", 16)
	namespace := envconf.RandomName("kind-ns", 16)

	testenv.Setup(
		envfuncs.CreateKindClusterWithConfig(kindClusterName, "kindest/node:v1.22.2", "kind-config.yaml"),
		envfuncs.CreateNamespace(namespace),
	)

	testenv.Finish(
		envfuncs.DeleteNamespace(namespace),
		envfuncs.DestroyKindCluster(kindClusterName),
	)
	os.Exit(testenv.Run(m))
}

func TestHelmChart(t *testing.T) {
           helmInfo: = {
			Name: "nginx",
                         Namespace: "default",
                         ReleaseName: "nginx-stable/nginx-ingress"
                         Version: "latest"
                 }
	tests := features.New("Setup Helm Chart").
                      SetupFromHelm(func(helmInfo Helm))
                      .Feature()
		
	test.Test(t, tests )
}
```

harshanarayana:
Closes #83

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,8 +55,12 @@ update-deps-go: ## Update all golang dependencies for this repo
 
 ##@ Tests
 
-.PHONY: test
-test: ## Runs golang unit tests
+.PHONY: install-helm test
+
+install-helm:  ## Install Helm toolchain for 3rd party integration
+	./hack/install-helm.sh
+
+test: install-helm ## Runs golang unit tests
 	./hack/test-go.sh
 
 ##@ Helpers
--- a/examples/third_party_integration/helm/helm_test.go
+++ b/examples/third_party_integration/helm/helm_test.go
@@ -1,0 +1,128 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package helm
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"sigs.k8s.io/e2e-framework/klient/k8s"
+	"sigs.k8s.io/e2e-framework/klient/wait"
+	"sigs.k8s.io/e2e-framework/klient/wait/conditions"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+	"sigs.k8s.io/e2e-framework/third_party/helm"
+)
+
+var curDir, _ = os.Getwd()
+
+func TestHelmChartRepoWorkflow(t *testing.T) {
+	feature := features.New("Repo based helm chart workflow").
+		Setup(func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
+			manager := helm.New(config.KubeconfigFile())
+			err := manager.RunRepo(helm.WithArgs("add", "nginx-stable", "https://helm.nginx.com/stable"))
+			if err != nil {
+				t.Fatal("failed to add nginx helm chart repo")
+			}
+			err = manager.RunRepo(helm.WithArgs("update"))
+			if err != nil {
+				t.Fatal("failed to upgrade helm repo")
+			}
+			err = manager.RunInstall(helm.WithName("nginx"), helm.WithNamespace(namespace), helm.WithReleaseName("nginx-stable/nginx-ingress"))
+			if err != nil {
+				t.Fatal("failed to install nginx Helm chart")
+			}
+			return ctx
+		}).
+		Assess("Deployment is running successfully", func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
+			deployment := &appsv1.Deployment{
+				ObjectMeta: v1.ObjectMeta{
+					Name:      "nginx-nginx-ingress",
+					Namespace: namespace,
+				},
+				Spec: appsv1.DeploymentSpec{},
+			}
+			err := wait.For(conditions.New(config.Client().Resources()).ResourceScaled(deployment, func(object k8s.Object) int32 {
+				return object.(*appsv1.Deployment).Status.ReadyReplicas
+			}, 1))
+			if err != nil {
+				t.Fatal("failed waiting for the Deployment to reach a ready state")
+			}
+			return ctx
+		}).
+		Assess("run Chart Tests", func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
+			manager := helm.New(config.KubeconfigFile())
+			err := manager.RunTest(helm.WithArgs("nginx"), helm.WithNamespace(namespace))
+			if err != nil {
+				t.Fatal("failed waiting for the Deployment to reach a ready state")
+			}
+			return ctx
+		}).
+		Teardown(func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
+			manager := helm.New(config.KubeconfigFile())
+			err := manager.RunRepo(helm.WithArgs("remove", "nginx-stable"))
+			if err != nil {
+				t.Fatal("cleanup of the helm repo failed")
+			}
+			return ctx
+		}).Feature()
+
+	testEnv.Test(t, feature)
+}
+
+func TestLocalHelmChartWorkflow(t *testing.T) {
+	feature := features.New("Local Helm chart workflow").
+		Setup(func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
+			manager := helm.New(config.KubeconfigFile())
+			err := manager.RunInstall(helm.WithName("example"), helm.WithNamespace(namespace), helm.WithChart(filepath.Join(curDir, "testdata", "example_chart")), helm.WithWait(), helm.WithTimeout("10m"))
+			if err != nil {
+				t.Fatal("failed to invoke helm install operation due to an error", err)
+			}
+			return ctx
+		}).
+		Assess("Deployment Is Running Successfully", func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
+			deployment := &appsv1.Deployment{
+				ObjectMeta: v1.ObjectMeta{
+					Name:      "example",
+					Namespace: namespace,
+				},
+				Spec: appsv1.DeploymentSpec{},
+			}
+			err := wait.For(conditions.New(config.Client().Resources()).ResourceScaled(deployment, func(object k8s.Object) int32 {
+				return object.(*appsv1.Deployment).Status.ReadyReplicas
+			}, 1))
+			if err != nil {
+				t.Fatal("failed waiting for the Deployment to reach a ready state")
+			}
+			return ctx
+		}).
+		Assess("run Helm Test Workflow", func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
+			manager := helm.New(config.KubeconfigFile())
+			err := manager.RunTest(helm.WithName("example"), helm.WithNamespace(namespace))
+			if err != nil {
+				t.Fatal("failed to perform helm test operation to check if the chart deployment is good")
+			}
+			return ctx
+		}).Feature()
+
+	testEnv.Test(t, feature)
+}
--- a/third_party/helm/helm.go
+++ b/third_party/helm/helm.go
@@ -1,0 +1,247 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package helm
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/vladimirvivien/gexe"
+	log "k8s.io/klog/v2"
+)
+
+type Opts struct {
+	// Name is used to indicate the name of the helm chart being processed
+	Name string
+	// Namespace is used to indicate the namespace in which the helm chart
+	// identified by Name will be processed
+	Namespace string
+	// ReleaseName is used to indicate the name of the release to be used
+	// while installing or upgrading the chart identified by Name into the
+	// Namespace
+	ReleaseName string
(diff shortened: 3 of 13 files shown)


support/helm/helm.go
```diff
@@ -0,0 +1,149 @@
...
+
+import (
+	"strings"
+
+	"github.com/vladimirvivien/gexe"
+	"github.com/vladimirvivien/gexe/exec"
+	log "k8s.io/klog/v2"
+)
+
+type Opts struct {
```
vladimirvivien, line 27: I like the option approach, will it cover most parameters for helm?

support/helm/helm.go
```diff
@@ -0,0 +1,149 @@
...
+	// chart in case if your tests require you to be able to deploy a packaged
+	// helm chart that is available locally as the tarball
+	Chart string
+	// Mode is used to indicate the mode in which the helm operation is being
+	// performed. These are the first level commands that are processed by the
+	// helm Binary
+	Mode string
+	// Args is used to pass any additional arguments that you might want to pass
+	// for running the helm command in question
+	Args []string
```
vladimirvivien, line 50: Field `Args` answers my question above.

support/helm/helm.go
```diff
@@ -0,0 +1,149 @@
...
+	command := m.getCommand(m.processOpts(opts...))
+	log.V(4).InfoS("Running Helm Operation", "command", command)
+	proc = m.e.RunProc(command)
+	return
+}
+
+func NewManager() *Manager {
+	mgr := &Manager{e: gexe.New()}
+	if mgr.e.Prog().Avail("helm") == "" {
+		panic(missingHelm)
```
vladimirvivien, line 146: No need to panic.
Since `Run` is the entry point, move the program check there.
Then return an error if prog is missing.

vladimirvivien (commented):
Good start!!!

support/helm/helm.go
crandles: I would add a `Wait bool`, `--wait` is pretty common thing to do with helm

harshanarayana: @crandles  I suppose `--wait` and `--timeout` both can be added as a direct helper calls. Let me do that.

support/helm/helm.go
harshanarayana: I have kept a set of options are directly invocable and other set of things as generic args so that we don't need to write a helper for each and every possible combination. I like @crandles  suggestion for wait being a direct invocable call for option. Will do that. Hopefully this generic `Args` should be enough for all other use

support/helm/helm.go
harshanarayana: All right. That makes sense. Let me change it accordingly.

vladimirvivien (commented):
This is looking good.
Please add `examples/helm` to show how to use this.

harshanarayana:
@vladimirvivien Safe to go by the expectation that the CICD infra would have the helm binary available or should I setup some way to get that installed first and run the test afterwards ?

support/helm/helm.go
```diff
@@ -0,0 +1,177 @@
...
+	command, err := m.getCommand(m.processOpts(opts...))
+	if err != nil {
+		return
+	}
+	log.V(4).InfoS("Running Helm Operation", "command", command)
+	proc = m.e.RunProc(command)
+	return
+}
+
+func NewManager() *Manager {
```
crandles, line 175: When you go to use this in an example I suspect you will want to  a way to pass in a kubeconfig reference, so that helm uses the right config generated by the test suite without an optional flag.

I am not sure if that should be required by default (e.g. in this initializer) or as an additional option.

harshanarayana: Yes. You are ready gut about that.. I am amusing then`WithArgs` to achieve that for now.. But I can extend the new object creation so that kube config is embedded into the helm object and avoid end user passing that one. Nice catch

vladimirvivien:
> @vladimirvivien Safe to go by the expectation that the CICD infra would have the helm binary available or should I setup some way to get that installed first and run the test afterwards ?

I am not sure if test-infra build environment already has helm as a binary.  Give it a try.

harshanarayana:
@vladimirvivien  Pushed the changes with the test. Can you give a go ahead for the tests to run ? Also, if the `helm` binaries are missing there do you have a preferred way to take care of that issue ? say adding a setup task that can just download the binary and place it in some dir ?

helayoty:
> 

I would say we should check if `helm` is installed or not as `kind` has already been handled [here](https://github.com/kubernetes-sigs/e2e-framework/blob/main/support/kind/kind.go#L164)

harshanarayana:
> > Safe to go by the expectation that the CICD infra would have the helm binary available or should I setup some way to get that installed first and run the test afterwards ?
> 
> @harshanarayana I would say we should check if `helm` is installed or not as `kind` has already been handled [here](https://github.com/kubernetes-sigs/e2e-framework/blob/main/support/kind/kind.go#L164)

@helayoty  Added a `hack/install-helm.sh` that checks and installs helm in CICD if required and the `helm.go` has a check to see if the `helm` is available or error out if not.

third_party/helm/helm.go
```diff
@@ -0,0 +1,208 @@
...
+		return
+	}
+	log.V(4).InfoS("Running Helm Operation", "command", command)
+	proc = m.e.RunProc(command)
+	return
+}
+
+func NewManager(kubeConfig string) *Manager {
+	return &Manager{e: gexe.New(), kubeConfig: kubeConfig}
+}
```
helayoty, line 208: How can we add/update the `helm` repo? 

```
helm repo add nginx-stable https://helm.nginx.com/stable
helm repo update
```

Makefile
```diff
@@ -57,6 +57,7 @@ update-deps-go: ## Update all golang dependencies for this repo
 
 .PHONY: test
 test: ## Runs golang unit tests
+	./hack/install-helm.sh
```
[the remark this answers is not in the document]
harshanarayana: Sure. Let me do that.

examples/third_party_integration/helm/helm_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"sigs.k8s.io/e2e-framework/third_party/helm"
+)
+
+var curDir, _ = os.Getwd()
+
+func TestHelmChartWorkflow(t *testing.T) {
+	feature := features.New("Helm chart workflow").
+		Setup(func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
+			manager := helm.NewManager(config.KubeconfigFile())
+			proc, err := manager.Run(helm.WithMode("install"), helm.WithName("example"), helm.WithNamespace(namespace), helm.WithChart(filepath.Join(curDir, "testdata", "example_chart")), helm.WithWait(), helm.WithTimeout("10m"))
```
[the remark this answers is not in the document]
harshanarayana: `WithReleaseName` is what will let you manage the one from the Repo and not in local. I can rename that call to make it `WithChartFromRepo` may be ?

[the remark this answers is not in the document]
harshanarayana: `manager.Run(helm.WithMode("install"), helm.WithName("example"), helm.WithNamespace(namespace), helm.WithReleaseName("nginx-stable/nginx-ingress"), helm.WithWait(), helm.WithTimeout("10m"))`

third_party/helm/helm.go
harshanarayana:
```go
manager.Run(helm.WithMode("repo"), helm.WithArgs("add", "nginx-stable", "https://helm.nginx.com/stable"))
manager.Run(helm.WithMode("repo"), helm.WithArgs("update"))
```

This should let you run that workflow. However, I can write a simple wrapper around this to make the workflow much easier. However,. Do you have a list of these common invocation that can benifit from having a wrappers ?

helayoty: No need to write a wrapper for common ones, just wanted to make sure we have this feature.

examples/third_party_integration/helm/helm_test.go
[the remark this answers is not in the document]
helayoty: Then, can we remove the example `helm` chart [here](https://github.com/harshanarayana/e2e-framework/tree/feature/git-83/enable-helm-support/examples/third_party_integration/helm/testdata/example_chart) and use one of the commonly used ones?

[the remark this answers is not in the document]
harshanarayana: I was worried about possible flaky tests and hence turned them into a local helm chart testdata. Should I not worry about it much and just use one of the standard helm repo chart ?

third_party/helm/helm.go
harshanarayana: Sure. The `WithMode` method is provided so that you can invoke any helm sub command and use it with `WithArgs` to pass any non standard args that are already provided with a helpers.

examples/third_party_integration/helm/helm_test.go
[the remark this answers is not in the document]
helayoty: I would suggest that. what do you think @vladimirvivien?

third_party/helm/helm.go
helayoty: Can we add this in the example?

examples/third_party_integration/helm/helm_test.go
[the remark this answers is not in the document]
harshanarayana: Or I can keep an example of both workflow as well. Or just turn the current chart into a `tgz` and avoid large number of files from getting added if that is another concern

third_party/helm/helm.go
harshanarayana: Sure. I am doing that as we speak to address the suggestion from https://github.com/kubernetes-sigs/e2e-framework/pull/90#discussion_r766907634

examples/third_party_integration/helm/helm_test.go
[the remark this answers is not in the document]
harshanarayana: Added an example of both now. Let me know if the local helm chart example needs to be cleaned up.

third_party/helm/helm.go
harshanarayana: Done.

third_party/helm/helm.go
```diff
@@ -0,0 +1,208 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/vladimirvivien/gexe"
+	"github.com/vladimirvivien/gexe/exec"
+	log "k8s.io/klog/v2"
+)
+
+type Opts struct {
```
vladimirvivien, line 28: Having tons of options can be useful, but I feel that this package is becoming a wrapper for the `helm` command. In that case, would it be better to just allow test writers to just pass in command in plain text to be easier ?

Having said that, my suggestion would be to select a very restricted set of functionalities that test users can do.  For instance, instead of allowing test writers to specify a mode in plain text, maybe this package should just support just 2 or 3 modes.

third_party/helm/helm.go
```diff
@@ -0,0 +1,208 @@
...
+// WithChart is used to configure the full path of the .tgz artifact of the Helm
+// chart in case if your tests require you to be able to deploy a packaged
+// helm chart that is available locally as the tarball
+func WithChart(chart string) Option {
+	return func(opts *Opts) {
+		opts.Chart = chart
+	}
+}
+
+// WithMode is used to invoke the first sub command of the helm binary. This can
```
vladimirvivien, line 112: As mentioned earlier, I don't think mode should be open-ended.  The API should have specific modes supported when creating tests.  I would suggest that the supported modes should be exposed as a method in the API as follows

```go
helm.New(configFile).RunRepo(WithName(...), WithArgs(...), etc)
helm.New(configFile).RunInstall(WithName(...), WithArgs(...), etc)
helm.New(configFile).RunTest(WithName(...), WithArgs(...), etc)
```

third_party/helm/helm.go
```diff
@@ -0,0 +1,208 @@
...
+	command, err := m.getCommand(m.processOpts(opts...))
+	if err != nil {
+		return
+	}
+	log.V(4).InfoS("Running Helm Operation", "command", command)
+	proc = m.e.RunProc(command)
+	return
+}
+
+func NewManager(kubeConfig string) *Manager {
```
vladimirvivien, line 206: Suggestion: Consider using just `New`

third_party/helm/helm.go
```diff
@@ -0,0 +1,208 @@
...
+	if opt.Timeout != "" {
+		commandParts = append(commandParts, "--timeout", opt.Timeout)
+	}
+	commandParts = append(commandParts, "--kubeconfig", m.kubeConfig)
+	return strings.Join(commandParts, " "), nil
+}
+
+// Run method is used to invoke a helm command to perform a suitable operation.
+// Please make sure to configure the right Opts using the Option helpers
+func (m *Manager) Run(opts ...Option) (proc *exec.Proc, err error) {
```
vladimirvivien, line 192: Avoid leaking type `*exec.Proc` which would force tests to depend on `gexe`. Instead, handle all errors, using gexe.Proc (i.e. proc.Error, proc.IsSuccess, etc) within the confined of the Run method and only return an error if something goes wrong.

vladimirvivien (commented):
@harshanarayana We are almost there.
Great progress. Hope my comments make sense.

examples/third_party_integration/helm/testdata/example_chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
+{{/*
```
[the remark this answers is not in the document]
harshanarayana: The charts were generated using helm create bootstrap. I did not customize anything other than some basic cleanup. But I don't see any reason why I can't just simplify this chart. Let me make this as primitive as possible and get rid of all the bells and whistles that helm added for it.

third_party/helm/helm.go
harshanarayana: Consider it done.

third_party/helm/helm.go
harshanarayana: The idea behind this was to help user/test workflow where they might want to log the outputs. But I can also take the approach of `log.V(4).Info` and dump it here and return the exit codes alone. Let me take care of it. Thanks

third_party/helm/helm.go
harshanarayana: > would it be better to just allow test writers to just pass in command in plain text to be easier

I thought about this. But if the arguments are getting programatically generated during the test, the builder pattern nature that exists now will make it more cleaner and handy to implement the command.  

That said, I do agree the options can get overwhelming. The current options provided as very minimal and used in almost all helm commands to get things done. However, I am fine removing the `WithMode` in favor of a few selective helm commands. 

1. Repo Add
2. Install Charts
3. Render Charts
4. Delete Charts
5. Run Tests

These are the most primitive one I can think of. Anything else I should consider or should I even remove some from the above list ?

harshanarayana:
@vladimirvivien  @helayoty I have updated the PR according to both of your suggestions. Please let me know if this looks all right. I have limited the helm operations to a minimal subset that are pre-defined unlike before

vladimirvivien (commented):
@harshanarayana This looks good to me.
@helayoty Are you ok with how this is working now?
I think it's ready for merge.

vladimirvivien (approved):
Thanks @harshanarayana for the great contribution.

/LGTM

PR #95, opened 2021-12-16
match on exact cluster names

crandles:
The current implementation of how we check for cluster existence is too permissive:

https://github.com/kubernetes-sigs/e2e-framework/blob/fee1391aeccdc260069bd5e0b25c6b187c2293c4/support/kind/kind.go#L87-L90

If clusters are created using a similar prefix or using a name that happens to be a substring of another cluster, this will not work as intended.

Adjusted the function signature slightly, in the attempt to reduce code duplication. _Most_ users should be unaffected as the addition is a variadic parameter:
```go
func (k *Cluster) CreateWithConfig(imageName, kindConfigFile string) (string, error) {
	return k.Create("--image", imageName, "--config", kindConfigFile)
}

func (k *Cluster) Create(args ...string) (string, error) {
...
```

### DIFF
--- a/support/kind/kind.go
+++ b/support/kind/kind.go
@@ -78,52 +78,43 @@ func (k *Cluster) getKubeconfig() (string, error) {
 	return file.Name(), nil
 }
 
-func (k *Cluster) CreateWithConfig(imageName, kindConfigFile string) (string, error) {
-	log.V(4).Info("Creating kind cluster ", k.name)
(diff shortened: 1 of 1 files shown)


ShwethaKumbla (approved):

cpanato (approved):
/lgtm

vladimirvivien (approved):
/lgtm

PR #102, opened 2022-01-21
Add *testing.T to before/after feature hooks

### LINKED ISSUE #101: Question: Can features pass/fail?
I was working on a plugin for someone and they have multiple features run in a single go test.

As a result, AfterEachTest gets hit only 5 times despite having 12 features.

What I'd like to do is be able to say `AfterEachFeature` (which I can do) and record each of those myself as a separate test result. However, there is nothing explicitly about pass/fail/skip status of the Feature type.

Is the idea of running multiple features in a single test opposed to the model e2e-framework wants us to use?

johnSchnake:
In order to get test pass/fail information in these hooks
we need access to the testing object.

Fixes #101 

** NOTE **
I wanted to add a test for this but the problem is that it passes the testing.T object around so even when we get the expected behavior, when we fail the test and confirm it cascades, there is no way to avoid `go test` from still counting that as a failure.

The workaround for this would be to make a mock that implements `testing.TB` and use that, but our signatures everywhere use *testing.T explicitly so that wont work.

Manually confirmed it did cascade though.

### DIFF
--- a/pkg/internal/types/types.go
+++ b/pkg/internal/types/types.go
@@ -33,7 +33,9 @@ type EnvFunc func(context.Context, *envconf.Config) (context.Context, error)
 // can be used to customized the behavior of the
 // environment. Changes to context are expected to surface
 // to caller. Meant for use with before/after feature hooks.
-type FeatureEnvFunc func(context.Context, *envconf.Config, Feature) (context.Context, error)
+// *testing.T is provided in order to provide pass/fail context to
+// features.
+type FeatureEnvFunc func(context.Context, *envconf.Config, *testing.T, Feature) (context.Context, error)
 
 // TestEnvFunc represents a user-defined operation that
 // can be used to customized the behavior of the
--- a/pkg/env/action.go
+++ b/pkg/env/action.go
@@ -71,7 +71,7 @@ func (a *action) runWithT(ctx context.Context, cfg *envconf.Config, t *testing.T
 }
 
 // runWithFeature will run the action and inject a FeatureInfo object into the callback function.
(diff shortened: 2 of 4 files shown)


pkg/internal/types/types.go
```diff
@@ -33,7 +33,9 @@ type EnvFunc func(context.Context, *envconf.Config) (context.Context, error)
 // can be used to customized the behavior of the
 // environment. Changes to context are expected to surface
 // to caller. Meant for use with before/after feature hooks.
-type FeatureEnvFunc func(context.Context, *envconf.Config, Feature) (context.Context, error)
+// *testing.T is provided in order to provide pass/fail context to
+// features.
+type FeatureEnvFunc func(context.Context, *envconf.Config, *testing.T, Feature) (context.Context, error)
```
vladimirvivien, line 38: I am wondering if something similar is needed for `StepFunc` where Assessments receive a copy of the currently executing `testing.T` context.

vladimirvivien (commented):
This looks good. Left comment.
Wonder if StepFunc should be extended to receive testing.T.

pkg/internal/types/types.go
johnSchnake: I don't mind taking those up if needed but I will say that is just _slightly_ tangential since this was necessary as part of the before/after hooks. The step funcs dont have hooks like those but it is possible a step would want to just abort if the test is already failed.

I'll file an issue for that and address it separately if thats OK?

vladimirvivien: @johnSchnake ok cool. No worries, that can come at a later time.

vladimirvivien (approved):
Thanks @johnSchnake 

/LGTM

PR #110, opened 2022-02-12
[WIP] Constructor api poc

vladimirvivien:


klient/constructor/deployment/constructor.go
```diff
@@ -0,0 +1,66 @@
...
+	return Constructor{
+		deployment: appsV1.Deployment{
+			ObjectMeta: deploymentMeta.Build(),
+			Spec: appsV1.DeploymentSpec{
+				Replicas: replicas,
+				Selector: &depSel,
+				Template: v1.PodTemplateSpec{
+					ObjectMeta: tempSpec.ObjectMeta,
+					Spec:       tempSpec.Spec,
+				},
```
harshanarayana, line 50: Can we not just use the `tempSpec` from line 39 ?
```suggestion
				Template: tempSpec,
```

klient/constructor/deployment/constructor.go
```diff
@@ -0,0 +1,66 @@
...
+// Build is the finalizer method that constructs and return a value of type appsV1.Deployment
+func (c Constructor) Build() appsV1.Deployment{
+	return c.deployment
+}
+
+// Replicas is an initializer func that converts int -> *int32
+func Replicas(r int) *int32 {
+	rep := int32(r)
+	return &rep
+}
```
harshanarayana, line 66: Can this be turned into a API under `Constructor` as well or for now we just want to keep the `Deployment` constructor as simple as possible with no major additional APIs ?

klient/constructor/meta/meta_constructor.go
```diff
@@ -0,0 +1,68 @@
...
+    ObjectMetaNone = ObjectMetaConstructor{obj: metaV1.ObjectMeta{}}
+)
+
+// ObjectMetaConstructor constructor for type coreV1.ObjectMeta
+type ObjectMetaConstructor struct {
+	obj metaV1.ObjectMeta
+}
+
+// Object is the initializer function for ObjectMetaConstructor
+func Object(name string) ObjectMetaConstructor {
```
harshanarayana, line 37: Can we provide some kind of support for `GenerateName` 

If we turn the `Object` into a no-args constructor, then we can provide a `.Name` and `.GenerateName` chained API calls to let user decide what they want to use ? It comes really handy to be able to generate pod names and resource names in cases where you are generating a bunch of pods/resources to load up the system for some test and you don't really care much about the name of each of them.

klient/constructor/meta/selector_constructor.go
```diff
@@ -0,0 +1,56 @@
...
+	sel metaV1.LabelSelector
+}
+
+// MatchLabels initializer function for type LabelSelectorConstructor
+func MatchLabels(labels map[string]string) LabelSelectorConstructor {
+	return LabelSelectorConstructor{sel: metaV1.LabelSelector{MatchLabels: labels}}
+}
+
+// MatchExpressions initializer function for type LabelSelectorConstructor
+func MatchExpressions(expressions...metaV1.LabelSelectorRequirement) LabelSelectorConstructor {
```
harshanarayana, line 37:
```suggestion
func MatchExpressions(expressions ...metaV1.LabelSelectorRequirement) LabelSelectorConstructor {
```

klient/constructor/meta/selector_constructor.go
```diff
@@ -0,0 +1,56 @@
...
+func (c LabelSelectorConstructor) MatchLabels(labels map[string]string) LabelSelectorConstructor {
+	c.sel.MatchLabels = labels
+	return c
+}
+
+// MatchExpressions setter for metaV1.LabelSelectorRequirement
+func (c LabelSelectorConstructor) MatchExpressions(expressions...metaV1.LabelSelectorRequirement) LabelSelectorConstructor {
+	c.sel.MatchExpressions = expressions
+	return c
+}
```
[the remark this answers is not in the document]
harshanarayana: What might be great eventually is providing a way to define the strategy of mutating values getting changed via the Fluent APIs. 

When an object is getting constructed, we default to mutate strategy as `Override` but if the users opt for `Update` we can update the values existing in the object as part of the API instead of overriding.  Am I overthinking this or is this useful to have ?

klient/constructor/container/constructor.go
```diff
@@ -0,0 +1,54 @@
...
+type Constructor struct {
+	container coreV1.Container
+}
+
+// Name initializer function for *Constructor
+func Name(name string) Constructor {
+	return Constructor{container: coreV1.Container{Name: name}}
+}
+
+// Image setter for container image name
```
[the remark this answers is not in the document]
harshanarayana: Also `volumeMounts` support

klient/constructor/pod/spec_constructor.go
```diff
@@ -0,0 +1,46 @@
...
+	spec := SpecConstructor{spec: coreV1.PodSpec{}}
+	for _, constructor := range containerConstructors {
+		spec.spec.Containers = append(spec.spec.Containers, constructor.Build())
+	}
+	return spec
+}
+
+// AddVolume is a setter method to store coreV1.Volume definition
+func (c SpecConstructor) AddVolume(vol coreV1.Volume) SpecConstructor {
+	c.spec.Volumes = append(c.spec.Volumes, vol)
```
harshanarayana, line 39: We might have to provide the companion API for this under containers constructor for `volumeMounts` right ?

harshanarayana (commented):
This is coming along wonderfully.

I do have a few general things that I think might be wonderful to have. 

1. You mentioned a great feature of being able to provide defaults where possible, but how do we go about doing them? Can we let the end users of the framework define certain details of resources that are namespaced vs cluster scoped and when we invoke the `Build` finalizers, we can leverage those to defaulting certain values. For ex. Being able to set resource values per namespace, or setting service accounts per namespace etc. 
```golang
pod.SetDefaults(pod.Defaulter{
    ServiceAccountName: "test-account",
    Resources: v1.ResourceList {
     ......
    } 
}, pod.NamespacedDefaults {
  "ns-1": pod.Defaulter {
   .....
   }
}
)
```
2. Would it make sense to perform some basic checks during the finalizer calls for some entities ? Say you have a VolumeMount created at the container level but missing a Volume at the Pod Spec level. This again is a very minor thing that is good to have and not something we might have to get done right away. 
3. Is there a way to automate generating some of these helper APIs for base entities and provide a top level wrapper that is handwritten so that we can support a good amount of options of the base entities via the fluent-apis ? Again this is another thing we can table for now and take a look at on a later date. Just dropping it here before the thought fades from my mind. 
4. In general the APIs this PR has right now are by default override in nature. You can't invoke them multiple times (not that this is a problem) but would it make sense to eventually turn these APIs into mutating in nature where they can support updating the existing values as well as override them ?


I will stop here for now. Let me run this thing locally once and re-formulate some thoughts and get back to this thread again. 

Thanks again for a wonderful PR @vladimirvivien  and sorry for the really lengthy thread.

klient/constructor/container/constructor.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// Commands setter method for container commands
+func (c Constructor) Commands(cmds...string) Constructor {
+	c.container.Command = cmds
+	return c
+}
+
+// Build finalizer method that returns the built *coreV1.Container
+func (c Constructor) Build() coreV1.Container {
```
crandles, line 52: It seems like this function always returns the internal object, and doesn't have much logic. Could the inner "object" being constructed be exposed instead of this function? or Perhaps we want a copy?

klient/constructor/deployment/strategy_constructor.go
```diff
@@ -0,0 +1,53 @@
...
+	StrategyRecreate = StrategyDefault
+)
+
+// StrategyConstructor is type to build values of type appsV1.DeploymentStrategy
+type StrategyConstructor struct {
+	strat appsV1.DeploymentStrategy
+}
+
+// StrategyRollingUpdate is an initializer function that creates a value of type appsV1.DeploymentStrategy
+func StrategyRollingUpdate(maxUnavailable string, maxSurge string) StrategyConstructor {
```
crandles, line 36: What if instead this was a "deployment mutator" -- you could remove the Constructor layer:
```go
func StrategyRollingUpdate(maxUnavailable string, maxSurge string) func(*appsv1.Deployment) error {
...
}
```

klient/constructor/pod/spec_constructor.go
```diff
@@ -0,0 +1,46 @@
...
+
+// Spec initializer method for type PodSpecConstructor
+func Spec(containerConstructors... container.Constructor) SpecConstructor {
+	spec := SpecConstructor{spec: coreV1.PodSpec{}}
+	for _, constructor := range containerConstructors {
+		spec.spec.Containers = append(spec.spec.Containers, constructor.Build())
+	}
+	return spec
+}
+
```
helayoty, line 36: I'd love if we can support `pod affinity` or `affinity` in general.

klient/constructor/deployment/constructor.go
helayoty: what will happen if one of the following builds fails? Do we clean up in case of one of the `build()` blocks falis? 
```
tempSpec := template.Build()
depSel := selector.Build()
ObjectMeta: deploymentMeta.Build()
```

klient/constructor/deployment/constructor.go
```diff
@@ -0,0 +1,66 @@
...
+		deployment: appsV1.Deployment{
+			ObjectMeta: deploymentMeta.Build(),
+			Spec: appsV1.DeploymentSpec{
+				Replicas: replicas,
+				Selector: &depSel,
+				Template: v1.PodTemplateSpec{
+					ObjectMeta: tempSpec.ObjectMeta,
+					Spec:       tempSpec.Spec,
+				},
+				Strategy: strategy.Build(),
```
helayoty, line 51: Is there a way to design one `build()` aggregator by the end of the whole object? i.e. check if `tempSpec` is not empty, then call `template.Construct.Build` internally?

klient/constructor/deployment/constructor.go
```diff
@@ -0,0 +1,66 @@
...
+	"sigs.k8s.io/e2e-framework/klient/constructor/meta"
+	"sigs.k8s.io/e2e-framework/klient/constructor/pod"
+)
+
+type Constructor struct {
+	deployment appsV1.Deployment
+}
+
+// Deployment is the initializer function for type deployment.Constructor
+func Deployment(
```
helayoty, line 32: I'm trying to think out loud about how we can scale this model without the need to write a 1:1. Is it possible to have a general constructor that takes kind/type? that will cover Deployment, StatefulSet, and DaemonSet. wdyt?

helayoty (commented):
Thank you very much for this wonderful effort. This will save a lot of time, lines of redundant code and will give the user much flexibility and consistency while writing the tests. 

I've added a few comments while trying to find a way to generalize the solution you've put together for more use cases. 

Also, I believe RBAC, configMap, persistence volumes, and secret store are key objects that many users would need frequently.

Great start and please let me know if I can help in this great effort.

klient/constructor/pod/spec_constructor.go
harshanarayana: I totally missed this one.  Even to this day, I still get confused with regards to the names and values to use for these. Will be great to provide a easy to use API for this. 🙏

vladimirvivien:
Thank you everyone for the thorough review.
I have been super busy and have not had time to dedicate to this.
But hopefully will resume on it in another week or so.

vladimirvivien:
@helayoty @crandles @cpanato @harshanarayana 
Apologies for neglecting this PR, especially after your insightful reviews.
I am rebooting this effort to get it done, hopefully by the next release.

klient/constructor/container/constructor.go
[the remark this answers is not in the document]
vladimirvivien: @harshanarayana the plan is to provide fluent API access to the entire Kubernetes Workload API objects.

klient/constructor/deployment/constructor.go
vladimirvivien: @helayoty there should not be any failure.
The `Build` step basically returns the object graph.

klient/constructor/deployment/constructor.go
vladimirvivien: This can be placed in a centralized API where it can be reused.

klient/constructor/meta/meta_constructor.go
vladimirvivien: @harshanarayana Yes. Ultimately, the API should provide similar behavior as what the API server does with defaults. Such as generate names (where appropriate) and provide sensible defaults where needed.

klient/constructor/meta/selector_constructor.go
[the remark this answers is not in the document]
vladimirvivien: @harshanarayana I would love to hear more about mutation vs reassignment of values. Keep in mind that this API is a one-shot API that builds the object graph. It is not meant to deal with state of objects after it's created.

klient/constructor/pod/spec_constructor.go
vladimirvivien: Yes. In the end, i'd like to have the entire workload API covered.

Issue #113, opened 2022-02-15
Automate release with GitHubActions


vladimirvivien:
Create GitHub Actions + Go releaser to automate releases.

cpanato:
@vladimirvivien can i take this one?

vladimirvivien:
@cpanato sure thing.

PR #115, opened 2022-02-16
add github actions/goreleaser to release the lib

cpanato:
Fixes: #113 

rehearsal on my fork: https://github.com/cpanato/e2e-framework/runs/5220056082?check_suite_focus=true
release: https://github.com/cpanato/e2e-framework/releases/tag/v99.99.01

/assign @vladimirvivien

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,33 @@
+name: release
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17.x
+
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          distribution: goreleaser
+          version: latest
+          args: release --rm-dist
+        env:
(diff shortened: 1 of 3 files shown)


cpanato:
if there is anything else to add let me know

.github/workflows/release.yml
```diff
@@ -0,0 +1,32 @@
+name: release
+
+on:
+  create:
```
vladimirvivien, line 4: What's the difference between `create:` and `push:`

vladimirvivien, line 4: Should this be `push:` instead of `create:`

.github/workflows/release.yml
```diff
@@ -0,0 +1,32 @@
...
+    tags:
+
```
vladimirvivien, line 6: Also, you may have to provide a tag name pattern for it to trigger (i.e. `v*` or `v*.*.**`)

vladimirvivien (approved):
Thanks for doing this! Left couple comments.

.github/workflows/release.yml
cpanato: both works for tags in this case, I can change for push, if you prefer

.github/workflows/release.yml
cpanato: makes sense, i will update, thanks for the review

vladimirvivien:
@cpanato I pressed `approved` by accident. Please see comments.

cpanato:
put a hold @vladimirvivien then i can follow up :)

.github/workflows/release.yml
cpanato: updated to push

vladimirvivien (approved):
/lgtm

Issue #112, opened 2022-02-15
[query/bug] mark testcase as failed and stop execution when `t.FailNow()` is encountered in an assessment


divmadan:
[Release v0.0.5]

I have a requirement to stop the testcase execution if certain assertions during the course of the testcase fails. I thought that we could control this behaviour by using [t.FailNow](https://pkg.go.dev/testing#T.FailNow) but it just marks that assertion as `FAIL` (thus behaving as [t.Fail](https://pkg.go.dev/testing#T.Fail)) and continues execution at the next assertion/feature.

Looks like it's due to the local scope of the variable `t *testing.T` in the assertion definition. 

Is this the desired behaviour? How should I work out my requirement, workarounds or suggestions?

***

MWE:

```go
// example_test.go

package example

import (
	"context"
	"testing"

	"sigs.k8s.io/e2e-framework/pkg/envconf"
	"sigs.k8s.io/e2e-framework/pkg/features"
)

func TestExample(t *testing.T) {
	failFeature := features.New("fail-feature").
		Assess("1==2", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
			if 1 != 2 {
				t.Log("1 != 2")
				t.FailNow() // mark test case as failed here, don't continue execution
			} else {
				t.Log("1 == 2")
			}
			return ctx
		}).
		Assess("print", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
			t.Log("THIS LINE SHOULDN'T BE PRINTED")
			return ctx
		}).
		Feature()

	nextFeature := features.New("next-feature").
		Assess("print", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
			t.Log("THIS LINE ALSO SHOULDN'T BE PRINTED")
			return ctx
		}).
		Feature()

	testenv.Test(t, failFeature, nextFeature)
}
```

```go
// main_test.go

package example

import (
	"log"
	"os"
	"path/filepath"
	"testing"

	"sigs.k8s.io/e2e-framework/pkg/env"
)

var testenv env.Environment

func TestMain(m *testing.M) {
	// get kubeconfig
	home, err := os.UserHomeDir()
	if err != nil {
		log.Fatal("error while getting user home directory")
	}
	testenv = env.NewWithKubeConfig(filepath.Join(home, ".kube", "config"))

	os.Exit(testenv.Run(m))
}
```

```
➜ go test . -v
[...]

PR #116, opened 2022-02-21
GIT-62: enable dry-run support

### LINKED ISSUE #62: Dry run
Supply a dry-run flag so we can check which tests would be run given the current flags.

harshanarayana:
Closes #62

TODO:

- [ ] Enable Stat Gathering
- [ ] Include UT for Flags + Env for dry-run


harshanarayana:
This is a very early draft mode of the test stat gathering mode without making any changes to the external facing APIs exposed by the Framework.

I still need to sort out a small goofup for the stats collection mechanism.  The current implementation leverages the dry-run equivalent mode to collect the stats first and then process them as regular tests. It does so by tricking the `Run` method to run twice. 

However, this leads to a small issue. Since the tests are run from `testing.M` the base tests will get marked as `PASS` though they are not really being run.  I am wondering what would be the way to avoid this. For now, I am not entirely sure what would be the easiest thing to do without making too many code changes and keeping the backward compatibility intact. 

I am open to suggestion. 

cc @vladimirvivien  @johnSchnake

johnSchnake:
> Since the tests are run from testing.M the base tests will get marked as PASS though they are not really being run

I don't think that is a problem as long as the output specifies it is in a dry-run mode. It is the same way with the ginkgo/e2e tests; when you do dry-run, to get the actual list of tests targeted you look for the ones that passed.

pkg/env/env.go
```diff
@@ -323,13 +361,19 @@ func (e *testEnv) Run(m *testing.M) int {
 		panic("context not set") // something is terribly wrong.
 	}
 
+	if !e.gatherMode {
```
ShwethaKumbla, line 364: Might be a silly question, is this stats will be collected by default eventhogh developer is not explicitely enabling gatherMode var?
If so I am just curious to know the necessity of gatherStats().

pkg/env/env.go
```diff
@@ -387,38 +433,116 @@ func (e *testEnv) getFinishActions() []action {
...
-		}
-
-		// skip feature which does not match with --feature
-		if e.cfg.FeatureRegex() != nil && !e.cfg.FeatureRegex().MatchString(featName) {
-			t.Skipf(`Skipping feature "%s": name not matched`, featName)
+func (e *testEnv) runTests() bool {
+	return !e.cfg.DryRunMode() && !e.gatherMode
+}
+
+func (e *testEnv) shoulFeatureOrAssessmentBeRun(f types.Feature, a types.Step, assessmentIndex int) (skipped bool, message string) {
```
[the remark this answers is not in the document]
harshanarayana: Thanks. Let me fix this. Totally missed it.

pkg/env/env.go
```diff
@@ -41,11 +43,23 @@ type (
...
 type testEnv struct {
-	ctx     context.Context
-	cfg     *envconf.Config
-	actions []action
-	rnd     rand.Source
+	ctx        context.Context
+	cfg        *envconf.Config
+	actions    []action
+	rnd        rand.Source
+	gatherMode bool
```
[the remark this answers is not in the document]
harshanarayana: Right now, the `gatherMode` is just an internal attribute used to change the behavior of how certain function behave. This is used to gather and setup a context attribute that provides the information about total number of tests, features and assessments to be run and keeps updating after each of them have been completed. It is currently not tracking the `passed` or `failed` or `skipped` state. But can be extended to provide that information. I do like the idea of having this information in the context. Let me modify this code a bit to include those details as well.

pkg/env/env.go
harshanarayana: The current way I am trying to gather stats is a very hacky on at the best since I wanted to retain the compatibility of the APIs intact and did not want to introduce a new top level construct. 

So, what I do is, when the `Run` method is called, we set the gather mode and run the tests in gather mode. Which basically aggregates the test results and do a recursive call to the `Run` again but this time with `gatherMode = false` as the deferred operation. This ensures the tests get executed now in the right context. 

But you are right that I don't really need the `gatherStats` method to do the same. It started with a different thought in my mind when I introduced this function and did not cleanup yet.  I can achieve this entire flow without actually having that function as well.

pkg/env/env.go
vladimirvivien, line 440: This method looks a little convoluted. I think part of the issue is (maybe) because it is also collecting stats for summary (see my suggestion for removing stats for now). Also, I would suggest delaying making the decision, of figuring out whether an action/feature should be executed, at this point.  Instead, consider pushing that decision futrther down the execution stack [here](https://github.com/kubernetes-sigs/e2e-framework/blob/319e09e8b645bfbae7fc332a32588882e693c813/pkg/env/action.go#L94)

vladimirvivien (commented):
@harshanarayana This is a good start to get dryrun support.  I have some suggestions prior to doing a deeper dive into the code.

Suggestion 0
Add a docs/design markdown which captures the aesthetic of both the UX (flags) and the API how this would work. It does not have to be lengthy.  Show how dry-run would be invoked from the UX and what the output could look like. 

(I will have more to suggest here later, need to think about what output will look like and how to override the Go's test API generated output).

Suggestion 1
I think you have two features being implemented. One is the dry-run and the other is report summary. I would suggest that you consider implementing dry-run only in this PR.

Reporting summary is something that would be useful (there is an issue for that actually) whether the test is doing actually test or dry-run.  So, it should be considered to be its own PR.

Suggestion 2
I think the code for the dry-run could be much simpler.  I would suggest to check for dry-run at a lower-level of the call stack, maybe somewhere like [here](https://github.com/kubernetes-sigs/e2e-framework/blob/319e09e8b645bfbae7fc332a32588882e693c813/pkg/env/action.go#L94) 

Suggestion 3
Because the framework leans heavily on the `go test` tool, I think this feature should also consider what happens when user uses `--list` flag. That is a minor consideration, but it should be considered nevertheless.

Lastly,
As you mentioned, this feature is also impacted by what is considered to be a test unit (TestXXX function vs Feature vs Assessment). Depending on what the test unit is, it will impact this PR as well.   See https://github.com/kubernetes-sigs/e2e-framework/issues/112

harshanarayana:
@vladimirvivien Thanks a lot for the detailed review. Let me address those changes and I am fine with splitting this PR into two different items. Should I create a new issue for that ?

pkg/env/env.go
```diff
@@ -387,38 +433,116 @@ func (e *testEnv) getFinishActions() []action {
...
+				skipped = true
+				message = fmt.Sprintf(`Skipping feature "%s": matched label provided in --skip-lables "%s=%s"`, name, k, f.Labels()[k])
+				return
 			}
 		}
+	}
+	return
+}
+
+func (e *testEnv) updateTestStats(m *testing.M, kind string, mode string, count int) {
```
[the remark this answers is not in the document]
harshanarayana: Yes. I am fine with splitting this for sure. I was taking care of the stats with the same PR since @johnSchnake  has suggested we should include the stats as well when discussing the orginal thread that introduced the ask for a `dry-run` flag. But you are right. It gets complicated to review/merge as well if we keep these two things together. +1 to moving this to a different PR/

pkg/env/env.go
harshanarayana: Let me move the stats part of this out to a different PR. 

Let me refactor the other change like you suggested into the execution part. I will push an update for this in a day or so.

vladimirvivien:
@harshanarayana Let me know if you need additional review for this.

harshanarayana:
> @harshanarayana Let me know if you need additional review for this.

Sorry I've been a bit backlogged at work.. Will get to this this weekend. Will update and let you know for a rereview

vladimirvivien:
@harshanarayana  no worries. Thanks for your help.

harshanarayana:
@vladimirvivien  I have done the following with the updated copy of this PR 

1. Move the stats part out and keep this only for `--dry-run`
2. Add a design doc about the basics of how this works
3. Integrate this with the `-test.list`
4. Cleanup a few duplicate code

I will update this by EOD with some more tests and a dedicated example for the entire workflow as well. Let me know what you think of the new changes in the PR. I have marked the earlier comments as resolved as I;ve taken care of most of them.  Please feel free to add new items. Would be happy to address them

vladimirvivien:
@harshanarayana Ok I will take a look.

docs/design/dry-run-mode.md
```diff
@@ -0,0 +1,83 @@
...
+the framework specific benefits of how the Tests to be processed can be listed
+
+## `--dry-run` mode
+`e2e-framework` adds a new CLI flag that can be used while invoking the test called `--dry-run`. This works in conjunction with `test.list` to provide the following behavior. 
+
+1. When the `--dry-run` mode is invoked No Setup/Teardown workflows are processed
+2. Will display the Assessments as individual tests like they would be processed if not invoked with `--dry-run` mode
+3. Skip all pre-post actions around the Before/After Features or Before/After Tests
+
+When tests are invoked with `-test.list` argument, the `--dry-run` mode is automatically switched to enabled to make sure setup/teardown as well as the pore-post actions can be skipped.
```
vladimirvivien, line 30: Correct `pore-post`

pkg/env/action.go
```diff
@@ -52,6 +53,10 @@ type action struct {
 func (a *action) runWithT(ctx context.Context, cfg *envconf.Config, t *testing.T) (context.Context, error) {
 	switch a.role {
 	case roleBeforeTest, roleAfterTest:
+		if cfg.DryRunMode() {
+			klog.V(2).Info("Skipping execution of roleBeforeTest and roleAfterTest due to framework being in dry-run mode")
```
vladimirvivien, line 57: This message should probably also appear when the e2e-framework is starting to inform the user that before/after step functions will be skipped in dry-run mode.

pkg/env/env.go
```diff
@@ -431,29 +420,80 @@ func (e *testEnv) execFeature(ctx context.Context, t *testing.T, featName string
...
 	})
 
 	return ctx
 }
 
+// requireProcessing is used as a generic helper utility that can be used to decide if a feature or assessment
+// needs to be skipped or not. This will simplify the process of checking for these behavior during the dry-run
+// mode. This also cleans up the duplication of code we had in execFeature with regards to how the skip was being
+// detected for feature and assessment
+func (e *testEnv) requireProcessing(f types.Feature, a types.Step, i int) (skip bool, message string) {
```
vladimirvivien, line 443: Please document what `i` is or give it a more descriptive name

vladimirvivien: Or, instead of `i`, pass in a `defaultName` string that will be used if feature does not have a name.

vladimirvivien: Actually, I think you should consider breaking this into two methods:

* requireFeatureProcessing()
* requireStepProcessing()

vladimirvivien (commented):
Hi @harshanarayana , again thank you for such great effort.
I left some comments.

docs/design/dry-run-mode.md
harshanarayana: Done.

vladimirvivien (approved):
This looks great! Thanks @harshanarayana .

harshanarayana:
Oh no! such a terrible mistake of missing the boilerplate text. Let me add and push the changes. ! 🤦

harshanarayana:
@vladimirvivien I need a LGTM again. Sorry for the pain.

vladimirvivien (approved):
@harshanarayana no worries.

harshanarayana:
urhg. Seems like a broken linter behavior. 

```bash
pkg/env/env.go:497:2: naked return in func `requireProcessing` with 31 lines of code (nakedret)
	return
	^
```
xref: https://github.com/golangci/golangci-lint/issues/1317 and https://github.com/alexkohler/nakedret/issues/18

harshanarayana:
@vladimirvivien  This one need an LGTM again.

harshanarayana:
```bash
❯ hack/verify-golangci-lint.sh                                                                                                                                                                                                                                      
golangci/golangci-lint info checking GitHub for tag 'v1.44.0'
----- snipped logs
❯ echo $?                                                                                                                                                                                                                                                           
0
```
check seem to run fine locally.

PR #117, opened 2022-02-21
GIT-94: enable extended Feature API to provide named setup and teardown

### LINKED ISSUE #94: Feature Builder doesn't support adding Step by name
The `feature` builder has a nice set of API exposed to be able to build setup and tear down methods.  But they auto generate the name of the step that is being run.

```go
fmt.Sprintf("%s-setup", b.feat.name)
```
However, this can be a bit of a problem to debug things. There is a really useful builder method called `WithStep` that can let you customize the name of the step. However, the argument used `level` used for this is coming from `pkg/internal/types/level.go`

```go
// WithStep adds a new step that will be applied prior to feature test.
func (b *FeatureBuilder) WithStep(name string, level Level, fn Func) *FeatureBuilder {
	b.feat.steps = append(b.feat.steps, newStep(name, level, fn))
	return b
}
```

Which means we can't really use that or invoke it from anywhere outside. Which means the examples can't use that either or anything that injects `e2e-framework` as a dependency. Does it make sense to move some bits from `pkg/internal/types` to a reusable package ?

harshanarayana:
Closes #94

### DIFF
--- a/pkg/features/builder.go
+++ b/pkg/features/builder.go
@@ -46,20 +46,29 @@ func (b *FeatureBuilder) WithStep(name string, level Level, fn Func) *FeatureBui
 
 // Setup adds a new setup step that will be applied prior to feature test.
 func (b *FeatureBuilder) Setup(fn Func) *FeatureBuilder {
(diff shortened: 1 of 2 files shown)


cpanato (commented):
/ok-to-test

vladimirvivien (approved):
Thank you @harshanarayana for getting this done!

LGTM

Issue #97, opened 2022-01-11
k8s resource watcher implementation


ShwethaKumbla:
K8S object watchers are great functionality provided by k8s to get efficient change notifications on resources.
The events supported by these watchers are
1. ADD
2. MODIFY/UPDATE
3. DELETE

The idea here is to make developer implementation easier. Without knowing the resource core type of k8s objects they have to just register their actions/functions for respective watch events using the provision provided by this framework and to stay informed about when these events gets triggered, just use Watch() which resides inside klient/k8s/resources package.

## Proposal
Watch function accepts a `object ObjectList` as an argument. ObjectList type is used to inject the resource type in which Watch has to be applied.

`klient/k8s/resources/resources.go`
```go=
import (
    "sigs.k8s.io/controller-runtime/pkg/client"
    "k8s.io/apimachinery/pkg/watch"
)

func (r *Resources) Watch(ctx context.Context,object client.ObjectList, opts client.ListOptions) watch.Interface {
    cl, err := client.NewWithWatch(cfg, client.Options{})
    if err != nil {
        log.Println("error while creating a watcher client", err)
        return
    }
    
    watcher, err := cl.Watch(ctx, object, &ops) 
     if err != nil {
        log.Println("error while creating a watcher client", err)
        return nil, err
    }
    
    return watcher
}
```

Watch() in resources.go will return the `watcher` type which helps to call `InvokeEventHandler()`. InvokeEventHandler accepts `EventHandlerFuncs` which carries the user registered function sets.

file : klient/k8s/resources/watch.go

```go=
// InvokeEventHandler triggers the registered methods based on the event received for particular k8s resources.
func (watcher watch.Interface)InvokeEventHandler(f EventHandlerFuncs{}) {
    for {
	select {
	case event, ok := <-watcher.ResultChan():
            // retrieve the event type
            eventType := event.Type
        
            switch eventType {
            case watch.Added:
                f.Add(event.Object)
            case watch.Modified:
                f.Update(event.Object)
            case watch.Deleted:
                f.Delete(event.Object)
            }
        } 
        
    }
}


type EventHandlerFuncs struct {
	AddFunc    func(obj interface{})
	UpdateFunc func(obj interface{})
	DeleteFunc func(obj interface{})
}

func (e EventHandlerFuncs) Add(obj interface{}) {
	...
}

func (e EventHandlerFuncs) Update(newObj interface{}) {
	...
}


[...]

PR #125, opened 2022-03-31
implement k8s resource watch and trigger action based on the events

ShwethaKumbla:
This PR provides an implementation for #97.

### DIFF
--- a/examples/watch_resources/README.md
+++ b/examples/watch_resources/README.md
@@ -1,0 +1,54 @@
+# Watching for Resource Changes
+
+The test harness supports several methods for querying Kubernetes object types and watching for resource states. This example shows how to watch particular resource and how to register the functions to act upon based on the events recieved.
+
+
+# Watch for the deployment and triger action based on the event
+
+Watch has to run as goroutine to get the different events based on the k8s resource state changes.
+```go
+func TestWatchForResources(t *testing.T) {
+...
+dep := appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{Name: "watch-dep", Namespace: cfg.Namespace()},
+			}
+
+			// watch for the deployment and triger action based on the event recieved.
+			go cl.Resources().Watch(&appsv1.DeploymentList{}, &client.ListOptions{
+				FieldSelector: fields.OneTermEqualSelector("metadata.name", dep.Name),
+				Namespace:     dep.Namespace}, cl.RESTConfig()).WithAddFunc(onAdd).WithDeleteFunc(onDelete).Start(ctx)
+...
+}
+```
+
+# Function/Action definition and registering these actions
+
+```go
+// onAdd is the function executed when the kubernetes watch notifies the
+// presence of a new kubernetes deployment in the cluster
+func onAdd(obj interface{}) {
+	dep := obj.(*appsv1.Deployment)
+	depName := dep.GetName()
+	fmt.Printf("Dep name recieved is %s", depName)
+	if depName == "watch-dep" {
+		fmt.Println("Dep name matches with actual name!")
+	}
+}
+
+// onDelete is the function executed when the kubernetes watch notifies
+// delete event on deployment
+func onDelete(obj interface{}) {
+	dep := obj.(*appsv1.Deployment)
+	depName := dep.GetName()
+	if depName == "watch-dep" {
+		fmt.Println("Deployment deleted successfully!")
+	}
+}
+```
+
+The above functions can be registered using Register functions(WithAddFunc(), WithDeleteFunc(), WithUpdateFunc()) defined under klient/k8s/watcher/watch.go as shown in the example.
+
+# How to stop the watcher
+Create a global EventHandlerFuncs variable to store the watcher object and call Stop() as shown in example TestWatchForResourcesWithStop() test.
+
+Note: User should explicitly invoke the Stop() after the watch once the feature is done to ensure no unwanted go routine thread leackage.
--- a/examples/watch_resources/README.md
+++ b/examples/watch_resources/README.md
@@ -14,7 +14,7 @@ dep := appsv1.Deployment{
 			}
 
 			// watch for the deployment and triger action based on the event recieved.
-			go cl.Resources().Watch(&appsv1.DeploymentList{}, &client.ListOptions{
+			cl.Resources().Watch(&appsv1.DeploymentList{}, &client.ListOptions{
 				FieldSelector: fields.OneTermEqualSelector("metadata.name", dep.Name),
 				Namespace:     dep.Namespace}, cl.RESTConfig()).WithAddFunc(onAdd).WithDeleteFunc(onDelete).Start(ctx)
 ...
--- a/examples/watch_resources/watch_test.go
+++ b/examples/watch_resources/watch_test.go
@@ -1,0 +1,173 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package watch_resources
+
+import (
+	"context"
+	"testing"
+
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/e2e-framework/klient/k8s/resources"
+	"sigs.k8s.io/e2e-framework/klient/k8s/watcher"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+func TestWatchForResources(t *testing.T) {
+	watchFeature := features.New("test watcher").WithLabel("env", "dev").
+		Setup(func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			cl, err := cfg.NewClient()
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			dep := appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{Name: "watch-dep", Namespace: cfg.Namespace()},
+			}
+
+			// watch for the deployment and triger action based on the event recieved.
+			go cl.Resources().Watch(&appsv1.DeploymentList{}, resources.WithFieldSelector(labels.FormatLabels(map[string]string{"metadata.name": dep.Name}))).
+				WithAddFunc(onAdd).WithDeleteFunc(onDelete).Start(ctx)
+
+			return ctx
+		}).
+		Assess("create watch deployment", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			// create a deployment
+			deployment := newDeployment(cfg.Namespace(), "watch-dep", 1)
+			client, err := cfg.NewClient()
+			if err != nil {
+				t.Fatal(err)
+			}
+			if err := client.Resources().Create(ctx, deployment); err != nil {
+				t.Fatal(err)
+			}
+			return context.WithValue(ctx, "test-dep", deployment)
+		}).
+		Teardown(func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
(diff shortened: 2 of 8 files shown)


klient/k8s/watcher/watch.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// OnDelete calls DeleteFunc if it's not nil.
+func (e *EventHandlerFuncs) OnDelete(obj interface{}) {
+	if e.deleteFunc != nil {
+		e.deleteFunc(obj)
+	}
+}
+
+func (e *EventHandlerFuncs) RegisterAddFunc(addfn func(obj interface{})) *EventHandlerFuncs {
```
vladimirvivien, line 88: Suggestion: `SetAddFunc`

klient/k8s/watcher/watch.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// EventHandler can handle notifications for events that happen to a resource.
+//  * OnAdd is called when an object is added.
+//  * OnUpdate is called when an object is modified.
+//  * OnDelete will get the final state of the item if it is known.
+// Start will be waiting for the events notification which is responsible
+// for invoking the registered user defined functions
+type EventHandler interface {
+	OnAdd(obj interface{})
```
vladimirvivien, line 39: These methods, `OnAdd`, `OnDelete`, and `OnUpdate` is redundant with RegisterAdd, RegisterDelete, and RegisterUpdate methods.  There should only be one set. When called, they should just save the function to be called during event processing.

klient/k8s/watcher/watch.go
```diff
@@ -0,0 +1,101 @@
...
+	OnUpdate(newObj interface{})
+	OnDelete(obj interface{})
+	Start()
+}
+
+// Start triggers the registered methods based on the event received for
+// particular k8s resources
+func (e EventHandlerFuncs) Start() {
```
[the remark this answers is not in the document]
vladimirvivien: The API should probably offer a way to stop watching as well.

vladimirvivien (commented):
@ShwethaKumbla This is a great start. I left some comments. Let me know if you have questions.

klient/k8s/resources/resources.go
```diff
@@ -182,3 +183,17 @@ func (r *Resources) Label(obj k8s.Object, label map[string]string) {
 func (r *Resources) GetScheme() *runtime.Scheme {
 	return r.scheme
 }
+
+func (r *Resources) Watch(ctx context.Context, object cr.ObjectList, opts *cr.ListOptions, cfg *rest.Config) *watcher.EventHandlerFuncs {
+	cl, err := cr.NewWithWatch(cfg, cr.Options{})
+	if err != nil {
+		return nil
```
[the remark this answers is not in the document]
ShwethaKumbla: Updated as suggested

examples/watch_resources/README.md
```diff
@@ -0,0 +1,50 @@
...
+func TestWatchForResources(t *testing.T) {
+...
+dep := appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{Name: "watch-dep", Namespace: cfg.Namespace()},
+			}
+
+			// watch for the deployment and triger action based on the event recieved.
+			go cl.Resources().Watch(ctx, &appsv1.DeploymentList{}, &client.ListOptions{
+				FieldSelector: fields.OneTermEqualSelector("metadata.name", dep.Name),
+				Namespace:     dep.Namespace}, cl.RESTConfig()).SetAddFunc(onAdd).RegisterDeleteFunc(onDelete).Start()
```
harshanarayana, line 19: #nit,  I think you missed updating this to `SetDeleteFunc` from `RegisterDeleteFunc`

harshanarayana:
```suggestion
				Namespace:     dep.Namespace}, cl.RESTConfig()).SetAddFunc(onAdd). SetDeleteFunc(onDelete).Start()
```

examples/watch_resources/watch_test.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// onAdd is the function executed when the kubernetes watch notifies the
+// presence of a new kubernetes deployment in the cluster
+func onAdd(obj interface{}) {
+	dep := obj.(*appsv1.Deployment)
+	depName := dep.GetName()
+	fmt.Printf("Dep name recieved is %s", depName)
+	if depName == "watch-dep" {
+		fmt.Println("Dep name matches with actual name!")
```
harshanarayana, line 102: I might be nitpicking here but can we change this to using `klog` instead ?

klient/decoder/decoder.go
```diff
@@ -132,7 +132,7 @@ func DecodeEach(ctx context.Context, manifest io.Reader, handlerFn HandlerFunc,
 	return nil
 }
 
-// Decode a stream of  documents of any Kind using either the innate typing of the scheme.
+// DecodeAll is a stream of  documents of any Kind using either the innate typing of the scheme.
```
harshanarayana, line 135: 🖖

examples/watch_resources/watch_test.go
```diff
@@ -0,0 +1,114 @@
...
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			dep := appsv1.Deployment{
+				ObjectMeta: metav1.ObjectMeta{Name: "watch-dep", Namespace: cfg.Namespace()},
+			}
+
+			// watch for the deployment and triger action based on the event recieved.
+			go cl.Resources().Watch(ctx, &appsv1.DeploymentList{}, resources.WithFieldSelector(labels.FormatLabels(map[string]string{"metadata.name": dep.Name}))).
```
harshanarayana, line 46: It would be great to have an example of `Stop` handling too. Also, would it be a good idea to document and say we should explicitly invoke the `Stop` after the watch for the feature is done ? Otherwise this will leak a bunch of routines over the span of tests right ?

klient/k8s/watcher/watch.go
```diff
@@ -0,0 +1,113 @@
...
+
+	w, err := cl.Watch(e.Ctx, e.K8sObject, e.ListOptions)
+	if err != nil {
+		return err
+	}
+
+	// set watcher object
+	e.watcher = w
+
+	for event := range e.watcher.ResultChan() {
```
harshanarayana, line 66: > If an error occurs or Stop() is called, the implementation will close this channel and release any resources used by the watch.

The `ResultChan` will close if there is an error that occurs right ? Do we need a way to re-register the watcher in those cases? Something similar to `retry.OnError` helper of sort ?

klient/k8s/watcher/watch.go
```diff
@@ -0,0 +1,113 @@
...
+	return nil
+}
+
+// Stop triggers stopping a particular k8s watch resources
+func (e *EventHandlerFuncs) Stop() {
+	e.watcher.Stop()
+}
+
+// SetAddFunc used to set action on create event
+func (e *EventHandlerFuncs) SetAddFunc(addfn func(obj interface{})) *EventHandlerFuncs {
```
harshanarayana, line 98: All across the `wait.go` or `resources.go` we have used `WithXYZ` as the function naming format. Should we follow the same convention here ?

harshanarayana (commented):
Thanks a lot for this features @ShwethaKumbla . I left a few minor changes / comments. Mostly good to haves.

ShwethaKumbla:
@harshanarayana thanks for reviewing and providing valuable feedback. I will address and update PR.

klient/k8s/resources/resources.go
```diff
@@ -184,16 +189,19 @@ func (r *Resources) GetScheme() *runtime.Scheme {
 	return r.scheme
 }
 
-func (r *Resources) Watch(ctx context.Context, object cr.ObjectList, opts *cr.ListOptions, cfg *rest.Config) *watcher.EventHandlerFuncs {
-	cl, err := cr.NewWithWatch(cfg, cr.Options{})
-	if err != nil {
-		return nil
-	}
+func (r *Resources) Watch(ctx context.Context, object k8s.ObjectList, opts ...ListOption) *watcher.EventHandlerFuncs {
```
vladimirvivien, line 192: @ShwethaKumbla Let's remove ctx from the signature here and used it when needed in `Start(ctx)`

klient/k8s/resources/resources.go
```diff
@@ -184,16 +189,19 @@ func (r *Resources) GetScheme() *runtime.Scheme {
...
 	}
 
-	return watcher.New(w)
+	o := &cr.ListOptions{Raw: listOptions}
+
+	return &watcher.EventHandlerFuncs{
+		ListOptions: o,
+		K8sObject:   object,
+		Cfg:         r.GetConfig(),
+		Ctx:         ctx,
```
vladimirvivien, line 205: See comment above about context.

klient/k8s/watcher/watch.go
vladimirvivien: If an error occurs, let the user restart the watch instead of attempting to recover with a retry (I think).

vladimirvivien (commented):
Apologies for the late review.
This is making good progress. Left comments.

examples/watch_resources/watch_test.go
ShwethaKumbla: Yes, thanks for pointing out this. Added an example how to use Stop().

klient/k8s/watcher/watch.go
```diff
@@ -0,0 +1,112 @@
...
+// for invoking the registered user defined functions.
+// Stop used to stop the watcher.
+type EventHandler interface {
+	Start(ctx context.Context)
+	Stop()
+}
+
+// Start triggers the registered methods based on the event received for
+// particular k8s resources
+func (e *EventHandlerFuncs) Start(ctx context.Context) error {
```
vladimirvivien, line 51: One more suggestion, then this will be lgtm.
Now that context is passed into Start, it must be tested to make sure it's valid and that it has not been cancelled.

1) check to make sure it's not cancelled or errored
```go
func (e *EventHandlerFuncs) Start(ctx context.Context) error {
    if ctx.Err() != nil {
        return ctx.Err()
    }
...
}
```

klient/k8s/watcher/watch.go
```diff
@@ -0,0 +1,112 @@
...
+
+	w, err := cl.Watch(ctx, e.K8sObject, e.ListOptions)
+	if err != nil {
+		return err
+	}
+
+	// set watcher object
+	e.watcher = w
+
+	for event := range e.watcher.ResultChan() {
```
vladimirvivien, line 65: 2) here the loop needs to check to make sure the context has not been cancelled. One way to do this is to use a for-select loop block as shown.

Also, note I am suggesting that the event processing loop be wrapped in a goroutine.  I think it's better if Start automatically does that rather than the caller of Start.

```go
func (e *EventHandlerFuncs) Start(ctx context.Context) error {

    go func() {
        for {
            select {
                case <-ctx.Done() :
                    if ctx.Err() != nil {
                        return ctx.Err()
                    }
                    return nil
                case event := <- e.watcher.ResultChan():
                    // process event here
            }
        }
    }()

     return nil
...
}

**Review by vladimirvivien — COMMENTED** [review]
Ok, Almost there. Left suggestions.

**Comment by ShwethaKumbla** [comment]
@vladimirvivien sorry for the delayed update on PR. I have addressed the review comments. PTAL

File: examples/watch_resources/watch_test.go [hunk 15]
```diff
@@ -0,0 +1,171 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/e2e-framework/klient/k8s/resources"
+	"sigs.k8s.io/e2e-framework/klient/k8s/watcher"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+func TestWatchForResources(t *testing.T) {
+	watchFeature := features.New("test watcher").WithLabel("env", "dev").
```
vladimirvivien, line 35: Because method `Start` now does the event processing in a goroutine, it is important to ensure that the callback methods are firing. So, I am proposing updating the test as follows


```go
func TestWatchForResourcesWithStop(t *testing.T) {
    addWait := make(chan struct{})
    onAdd := func (obj interface{}) {
	dep := obj.(*appsv1.Deployment)
	depName := dep.GetName()
	if depName == "watch-dep"  {
		 addWait <- struct{}{}
	}
    }
    ...
    Assess("create watch...", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
        ...
	if err := client.Resources().Create(ctx, deployment); err != nil {
	    t.Fatal(err)
	}
        // After creation, wait for signal
        select {
            case <-time.After(3 *time.Second):
            t.Error("Add callback not called)
            case <-addWait:
            close(addWait)
        }
        ...
        return context.WithValue(ctx, "test-dep", deployment)
    })
}
```

You can also do something similar to watch for deleted messaging firing.

vladimirvivien (commented):
This is looking great. I left a comment about checking callback methods are firing in test.
I think after this we're good to merge.

vladimirvivien (approved):
Merging, changes to test will be handled in a different PR.

PR #129, opened 2022-04-07
Add support for cluster API objects

johnSchnake:
Added the 3 apis to the scheme used by the klient so that it knows
how to deal with those objects.

Added a test for this behavior and modified the cluster setup
to install clusterctl and use it to install cluster API.

Currently the test is gated by the GITHUB_TOKEN being set since
otherwise we run into rate limiting issues.


klient/k8s/resources/resources_test.go
```diff
@@ -237,3 +239,31 @@ func TestListAllPods(t *testing.T) {
 
 	t.Logf("pod list contains %d pods", len(pods.Items))
 }
+
+func TestListClusterAPI(t *testing.T) {
+	ghToken := os.Getenv("GITHUB_TOKEN")
+	if len(ghToken) == 0 {
+		t.Skip("Skipping clusterAPI test due to absence of github token. See issue https://github.com/kubernetes/test-infra/issues/25892 for details.")
```
johnSchnake, line 246: Just skips the test if GITHUB_TOKEN isnt set

klient/internal/testutil/setup.go
```diff
@@ -143,3 +158,28 @@ func waitForControlPlane(c kubernetes.Interface) error {
 		})
 	return err
 }
+
+// Ensures all pods for clusterAPI are running.
+func waitForClusterAPI(c kubernetes.Interface) error {
+	log.Info("Waiting for clusterAPI pods to be initialized...")
+	err := wait.Poll(5*time.Second, time.Minute*2,
+		func() (bool, error) {
+			pods, err := c.CoreV1().Pods("capi-system").List(context.TODO(), metav1.ListOptions{})
```
johnSchnake, line 167: The only pod is the capi-controller and if that isnt running then the webhook throws an error when the test tries to create a cluster object.

klient/k8s/resources/resources.go
```diff
@@ -179,3 +184,18 @@ func (r *Resources) Annotate(obj k8s.Object, annotation map[string]string) {
 func (r *Resources) Label(obj k8s.Object, label map[string]string) {
 	obj.SetLabels(label)
 }
+
+func getScheme() *runtime.Scheme {
```
vladimirvivien, line 188: I think hard-coding the Cluster-API schemes would create an explicit/forced dependency on CAPI in e2e-framework.  Instead, I'd like to propose the expansion of the `Resources` type in `klient/k8s/resources` to provide the ability to add schemes as needed. Maybe something like :

```go
type SchemeAdder func(s *Scheme) error

type Resources {
...
}

func (r *Resources) AddSchemes(adders... SchemeAdder) error {
    for _, add := range adders {
        if err := add(r.scheme); err != nil {
            return err
        }
    }
    return nil
}
```

Then, when using the Resources type, schemes can be added as shown below:

```go
func SomeTest(...) {
    res, err := New(cfg)
    if err != nil {
	t.Fatalf("Error creating new resources object: %v", err)
    }
    if err := res.AddSchemes(
        clusterapiv1beta1.AddToScheme, 
        clusterapiv1alpha3.AddToScheme,
        clusterapiv1alpha4.AddToScheme
    ); err != nil {
        t.Fatalf("failed to add scheme: %s", err)
    }
    ...
}
```

vladimirvivien: That way there is no additional dependencies (other than the standard API objects) added to the e2e-framework itself.

Thoughts?

vladimirvivien (commented):
Hi @johnSchnake Thanks for starting this.
I left a comment about expanding the Resources type to make it easy to add Schemes without adding direct CAPI dependencies to the e2e-framework itself.

klient/k8s/resources/resources.go
johnSchnake: As I was typing out a long response I convinced myself of something close to the above. How about this:

 - expose the scheme-adding capability (love that idea and hate I didnt already expose that)
 - Remove all the clusterctl dependency
 - add a test that registers another CRD (thats lighter-weight) and confirms the addingToScheme logic works the way we want
 - still add the clusterAPI types by default.

This would:
 - expose the functionality to cover more cases
 - test the functionality regardless of GITHUB_TOKEN/rate limit issues
 - remove all the clusterctl stuff
 - still support clusterAPI 'out of the box' without having to manually register those types.

So the dependency to clusterctl is there, but much lighter weight since we are just adding the go.mod dependency and not clusterctl installation etc (which I think is where the main burden comes from). In this way we set a precedent that if a certain CRD/project is popular enough we can automatically register the types, but for lesser used or custom ones we still have a support route.

What do you think of that? Or are you against any sort of binding to the cluster-api types?

vladimirvivien: @johnSchnake 
That sounds reasonable, I would have to see the code to completely understand your suggestion and we can iterate from there.  Thanks!

johnSchnake: Change made and I like this more too.
 - Adds method to get the current scheme; this allows you to add whatever you want to it.
 - Adds clusterAPI types by default
 - Adds a test (the bulk of the code) with a CRD, confirms the `Resource` object can't operate on it, register the type, and confirm it works. I wanted to make sure that after the resource object is already created and used that you can continue to modify the scheme.

vladimirvivien: @johnSchnake, I still don't think a default dependency on Cluster-API types is a good idea. Unless I am missing something it looks like function `defaultScheme()` is still pulling the cluster-api packages by default. 

Also, keep in mind that testing custom resources, like Cluster-API types, requires additional ceremonial steps like type registration and type definition (and possibly a controller). So, testing CRDs alone go beyond the vanilla testing most folks would write. That is why I think the test-user should add the scheme as needed.

Thoughts.

johnSchnake: > Unless I am missing something it looks like function defaultScheme() is still pulling the cluster-api packages by default.

It does; I had said that I was in support of making that dependence by default. If you feel strongly I can take it out.

> Also, keep in mind that testing custom resources, like Cluster-API types, requires additional ceremonial steps like type registration and type definition (and possibly a controller). So, testing CRDs alone go beyond the vanilla testing most folks would write. That is why I think the test-user should add the scheme as needed.

I'm not sure I understand fully what your message is here. For _our_ purposes we just need to make sure that we can handle the types (ie no controller but the types do need registered). Hence, for the test I have a custom type that I register with the server and add the schema to ensure all the bits work as expected. Are you fine with that part? If so I'll just remove the cluster-api dependency and those users still have a path to utilizing the e2e-framework klient with those custom types.

vladimirvivien: @johnSchnake 
To be clear, I support what you are doing (i.e. create a sample resource and running a test against it, to ensure custom types like CAPI, and their schemes, would work).

What I am saying is that beyond that, we probably should not inject a capi type dependency in the core of klient package directly. Instead, give test users a way to inject their types as needed during the test.

Hope that made sense.

johnSchnake:
Removed dependency on CAPI types and just used our test/example type for the test. We can point users to that test to show that it can be done if they want to use CAPI+e2e-framework

klient/k8s/resources/resources.go
```diff
@@ -179,3 +178,7 @@ func (r *Resources) Annotate(obj k8s.Object, annotation map[string]string) {
 func (r *Resources) Label(obj k8s.Object, label map[string]string) {
 	obj.SetLabels(label)
 }
+
+func (r *Resources) GetScheme() *runtime.Scheme {
```
johnSchnake, line 182: Just exposing it via a method since it had no other exported fields. This felt more consistent.

klient/k8s/resources/resources_test.go
```diff
@@ -237,3 +240,43 @@ func TestListAllPods(t *testing.T) {
...
+
+	// See that we can't list it because we don't know the type.
+	ps := &projectExample.ProjectList{}
+	err = res.List(context.TODO(), ps)
+	if err == nil {
+		t.Error("Expected error while listing custom resources before adding it to scheme, but got none")
+	}
+
+	// Register type with klient.
+	if err := projectExample.AddToScheme(res.GetScheme()); err != nil {
```
johnSchnake, line 273: Example of how you'd actually use this. If someone asks, just point them to this line.

vladimirvivien: Ok great, that will work too. Put the onus of scheme registration on tester.

vladimirvivien (approved):
LGTM
Thanks @johnSchnake !!

PR #135, opened 2022-05-02
GIT-109: enable CRD setup helper to ease the testing of Operators

### LINKED ISSUE #109: Add example of how to test a kubebuilder project
kubebuilder leverages [envtest](https://book.kubebuilder.io/reference/envtest.html) in the test suite it generates with its scaffold infrastructure for creating kubernetes controllers. `envtest` is not a full kubernetes cluster, so some complex controller implementations are unable to be tested using the tooling without mocks.

The existing envtest has a basic usage that should be easily supported by this framework:
```go
// By default, kubebuilder implements this (via scaffolding) using the Ginkgo suite.

import sigs.k8s.io/controller-runtime/pkg/envtest

//specify testEnv configuration
testEnv = &envtest.Environment{
    CRDDirectoryPaths: []string{filepath.Join("..", "config", "crd", "bases")},
}

//start testEnv
cfg, err = testEnv.Start()

//write test logic

//stop testEnv
err = testEnv.Stop()
```

To simplify usage, an `envfunc` helper (or just an implementation of the types.EnvFunc interface) should be added to simplify the crd install / uninstall (using the `decoder` package).

Alternatively: import and inter-op with the [envtest](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest) implementation installing CRDs and webhooks.

harshanarayana:
Closes #109 

cc @crandles  Hope this address the ask you have had in the Issue you opened. 

```bash
=== RUN   TestCRDSetup
=== RUN   TestCRDSetup/Custom_Controller
=== RUN   TestCRDSetup/Custom_Controller/Check_If_Resource_created
I0502 19:36:18.886351   72899 envtest_test.go:60] "CR Details" cr=&{TypeMeta:{Kind: APIVersion:} ObjectMeta:{Name:my-new-cron-object GenerateName: Namespace:my-ns-6dde SelfLink: UID:6769a909-6188-40a5-bc28-6e50df6a561b ResourceVersion:546 Generation:1 CreationTimestamp:2022-05-02 19:36:18 +0530 IST DeletionTimestamp:<nil> DeletionGracePeriodSeconds:<nil> Labels:map[] Annotations:map[] OwnerReferences:[] Finalizers:[] ClusterName: ManagedFields:[{Manager:envtest.test Operation:Update APIVersion:stable.example.com/v1 Time:2022-05-02 19:36:18 +0530 IST FieldsType:FieldsV1 FieldsV1:{"f:spec":{".":{},"f:cronSpec":{},"f:image":{}}} Subresource:}]} Spec:{CronSpec:* * * * */5 Image:my-awesome-cron-image Replicas:0}}
--- PASS: TestCRDSetup (2.08s)
    --- PASS: TestCRDSetup/Custom_Controller (2.08s)
        --- PASS: TestCRDSetup/Custom_Controller/Check_If_Resource_created (0.02s)
PASS
ok  	sigs.k8s.io/e2e-framework/examples/third_party_integration/envtest	45.813s
```

### DIFF
--- a/klient/decoder/decoder.go
+++ b/klient/decoder/decoder.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"os"
 	"strings"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -90,6 +91,20 @@ func DecodeAllFiles(ctx context.Context, fsys fs.FS, pattern string, options ...
 		return nil
 	}, options...)
 	return objects, err
+}
+
+// ApplyWithManifestDir resolves all the files in the Directory dirPath against the globbing pattern and creates a kubernetes
+// resource for each of the resources found under the manifest directory.
+func ApplyWithManifestDir(ctx context.Context, r *resources.Resources, dirPath, pattern string, createOptions []resources.CreateOption, options ...DecodeOption) error {
+	err := DecodeEachFile(ctx, os.DirFS(dirPath), pattern, CreateHandler(r, createOptions...), options...)
+	return err
+}
+
+// DeleteWithManifestDir does the reverse of ApplyUsingManifestDir does. This will resolve all files in the dirPath against the pattern and then
+// delete those kubernetes resources found under the manifest directory.
+func DeleteWithManifestDir(ctx context.Context, r *resources.Resources, dirPath, pattern string, deleteOptions []resources.DeleteOption, options ...DecodeOption) error {
+	err := DecodeEachFile(ctx, os.DirFS(dirPath), pattern, DeleteHandler(r, deleteOptions...), options...)
+	return err
 }
 
 // Decode a stream of documents of any Kind using either the innate typing of the scheme.
--- a/pkg/envfuncs/resource_funcs.go
+++ b/pkg/envfuncs/resource_funcs.go
@@ -1,0 +1,51 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package envfuncs
+
+import (
+	"context"
+
+	"sigs.k8s.io/e2e-framework/klient/decoder"
+	"sigs.k8s.io/e2e-framework/klient/k8s/resources"
+	"sigs.k8s.io/e2e-framework/pkg/env"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+)
+
+// SetupCRDs is provided as a helper env.Func handler that can be used to setup the CRDs that are required
+// to process your controller code for testing. For additional control on resource creation handling, please
+// use the decoder.ApplyWithManifestDir directly with suitable arguments to customize the behavior
+func SetupCRDs(crdPath, pattern string) env.Func {
+	return func(ctx context.Context, c *envconf.Config) (context.Context, error) {
+		r, err := resources.New(c.Client().RESTConfig())
+		if err != nil {
+			return ctx, err
+		}
+		return ctx, decoder.ApplyWithManifestDir(ctx, r, crdPath, pattern, []resources.CreateOption{})
+	}
+}
+
+// TeardownCRDs is provided as a handler function that can be hooked into your test's teardown sequence to
+// make sure that you can cleanup the CRDs that were setup as part of the SetupCRDs hook
+func TeardownCRDs(crdPath, pattern string) env.Func {
+	return func(ctx context.Context, c *envconf.Config) (context.Context, error) {
+		r, err := resources.New(c.Client().RESTConfig())
+		if err != nil {
+			return ctx, err
+		}
+		return ctx, decoder.DeleteWithManifestDir(ctx, r, crdPath, pattern, []resources.DeleteOption{})
+	}
+}
--- a/examples/crds/README.md
+++ b/examples/crds/README.md
@@ -1,0 +1,25 @@
+# Kubernetes Custom Resource Integration Test
+
+While developing a kubernetes operator for your custom use cases, it is very common practice that you end up creating your own custom resources.
+
+This example shows how to leverage the helper functions provided by the Framework itself to setup
+the CRD resources using the `decoder` package against your test cluster before starting the actual test workflow. 
+
+## How does this work ?
+
+1. You can leverage the framework's `env.Func` type helper for setting up the CRDs and tearing them down after the tests
+2. Register the CRD scheme with the `resources.Resources` to leverage the helpers for interacting with Custom resource objects
+
+## What does this test do ?
+
+1. Create a Kind cluster with a random name generated with `crdtest-` as the cluster name prefix
+2. Create a custom namespace with `my-ns` as the prefix
+3. Register the CRDs listed under `./testdata/crds` using the resource decode helpers
+4. Create a new Custom Resource for the CRD created in step #3
+5. Fetch the CR created in Test setup and print the value
+
+## How to run the tests
+
+```bash
+go test -v .
+```
--- a/examples/crds/envtest_test.go
+++ b/examples/crds/envtest_test.go
@@ -1,0 +1,65 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package crds
+
+import (
+	"context"
+	"os"
+	"testing"
+
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/e2e-framework/examples/crds/testdata/crontabs"
+	"sigs.k8s.io/e2e-framework/klient/decoder"
+	"sigs.k8s.io/e2e-framework/klient/k8s/resources"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+func TestCRDSetup(t *testing.T) {
+	feature := features.New("Custom Controller").
+		Setup(func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+			r, err := resources.New(c.Client().RESTConfig())
+			if err != nil {
+				t.Fail()
+			}
+			crontabs.AddToScheme(r.GetScheme())
+			r.WithNamespace(namespace)
+			decoder.DecodeEachFile(
+				ctx, os.DirFS("./testdata/crs"), "*",
+				decoder.CreateHandler(r),
+				decoder.MutateNamespace(namespace),
+			)
+			return ctx
+		}).
+		Assess("Check If Resource created", func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+			r, err := resources.New(c.Client().RESTConfig())
+			if err != nil {
+				t.Fail()
+			}
+			r.WithNamespace(namespace)
+			crontabs.AddToScheme(r.GetScheme())
+			ct := &crontabs.CronTab{}
+			err = r.Get(ctx, "my-new-cron-object", namespace, ct)
+			if err != nil {
+				t.Fail()
+			}
+			klog.InfoS("CR Details", "cr", ct)
+			return ctx
+		}).Feature()
+
+	testEnv.Test(t, feature)
+}
--- a/examples/crds/main_test.go
+++ b/examples/crds/main_test.go
@@ -1,0 +1,53 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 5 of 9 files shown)


examples/third_party_integration/envtest/README.md
```diff
@@ -0,0 +1,25 @@
+# EnvTest Integration
```
vladimirvivien, line 1: Let's title this `Kubernetes Custom Resource Integration Test`
This work really has nothing to do with envtest.
So, it would be better if this is titled exactly what it is so folks know what it does immediately.

examples/third_party_integration/envtest/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+`envtest` is a companion project to be used with `kubebuilder` based projects where you are creating your own controllers based on CRDs to enable quick test workflows.
```
vladimirvivien, line 3: Although this does something similar, in functionality,  as envtest, make sure to mention that this is not envtest and does not use it.

examples/third_party_integration/envtest/envtest_test.go
```diff
@@ -0,0 +1,65 @@
+/*
```
vladimirvivien, line 1: Since this is an example that is using the e2e-framework directly, I don't think it needs to go in `examples/third_party_integration/envtest` directory. This could just go in `examples/CRDs`

third_party/envtest/envtest.go
```diff
@@ -0,0 +1,61 @@
+/*
```
vladimirvivien, line 1: Again, since this has nothing to do with envtest, I would suggest to move it to `pkg/envfuncs` package because it makes sense as a core part of the framework.

third_party/envtest/envtest.go
```diff
@@ -0,0 +1,61 @@
...
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+)
+
+// SetupCRDs is provided as a helper env.Func handler that can be used to setup the CRDs that are required
+// to process your controller code for testing. This handlers provides a functionality similar to the one
+// provided by the envtest framework via the `CRDDirectoryPaths` argument of the envtest helper.
+//
+// The crdPath should be the base directory containing all the CRDs that needs to be pre-installed into
+// your cluster before your test handlers are kicked off. This can be integrated with the setup handlers
+func SetupCRDs(crdPath string) env.Func {
```
vladimirvivien, line 35: I would suggest to create an `ApplyYamlDir` function that test developers can use to apply any yamls (not just CRDs). Then you can create a `SetupCRDs` that calls the former.

third_party/envtest/envtest.go
```diff
@@ -0,0 +1,61 @@
...
+		err = decoder.DecodeEachFile(
+			ctx, os.DirFS(crdPath), "*", decoder.CreateHandler(r),
+		)
+		return ctx, err
+	}
+}
+
+// TeardownCRDs is provided as a handler function that can be hooked into your test's teardown sequence to
+// make sure that you can cleanup the CRDs that were setup as part of the SetupCRDs hook
+func TeardownCRDs(crdPath string) env.Func {
```
vladimirvivien, line 50: Similarly, I would suggest creating `DeleteYamlDir` function, that can be used in general. Then you can create a `TeardownCRDs` that calls the former.

vladimirvivien (commented):
@harshanarayana This is great and I like it.

Because this has nothing to do with envtest, the tool, I don't think you need to put the code and the example in `third_party_integration`.  This is basically using APIs already built-in the framework, so there is no need to put it there.

I left specific comments to guide you.

crandles:
This looks great -- FYI envtest includes some additional functionality, notably:

1. applying conversion webhook configuration to CRDs

```go
	if err := modifyConversionWebhooks(options.CRDs, options.Scheme, options.WebhookOptions); err != nil {
		return nil, err
	}
....
// modifyConversionWebhooks takes all the registered CustomResourceDefinitions and applies modifications
// to conditionally enable webhooks if the type is registered within the scheme.
```
https://github.com/kubernetes-sigs/controller-runtime/blob/v0.11.2/pkg/envtest/crd.go#L99-L101

2. installing webhooks

A custom webhook install process with options for modifying the cabundle at runtime.

It seems to do some subtle actions with port selection, that I am still unpacking.

https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest#WebhookInstallOptions

3. Waiting for CRDs/Webhooks

Both CRD and Webhook installers have an option to wait for readiness. This should be easy to add in with the conditions/wait packages.

---

What would you think of using [envtest.CRDInstallOptions](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest#CRDInstallOptions) as a parameter for the envfunc?

Additionally, I am worried about some of the subtle patching behavior that otherwise "just works". Another strategy would be to use `decoder` to load objects, and the `envtest` Install functions to execute.

third_party/envtest/envtest.go
```diff
@@ -0,0 +1,61 @@
...
+// The crdPath should be the base directory containing all the CRDs that needs to be pre-installed into
+// your cluster before your test handlers are kicked off. This can be integrated with the setup handlers
+func SetupCRDs(crdPath string) env.Func {
+	return func(ctx context.Context, c *envconf.Config) (context.Context, error) {
+		r, err := resources.New(c.Client().RESTConfig())
+		if err != nil {
+			return ctx, err
+		}
+		err = decoder.DecodeEachFile(
+			ctx, os.DirFS(crdPath), "*", decoder.CreateHandler(r),
```
crandles, line 42: If yaml files that are not CRDs are present in the directory, this will attempt to create those as well. This may not be expected -- I think we may want an additional decoder func (or option, TBD) to only apply the "handler" function on a given type.

third_party/envtest/envtest.go
crandles: those sound useful, and like things I have locally 😄 , though we may want CRD/envtest specific functions for some of the reasons I mention in this comment: https://github.com/kubernetes-sigs/e2e-framework/pull/135#issuecomment-1116645489

third_party/envtest/envtest.go
harshanarayana: I added a simple Pre Handler filter function option to the newly created method to retain the backward compatibility. We can decided how do we enable this for existing APIs that we already released into the wild

harshanarayana:
@crandles 

Not sure if all the features in the `envtest` are something we should worry about supporting here one to one. There are helpers here that can do most of them with a bit of extra code. Not sure if we can write up a generic mechanisms that can support workflow such as 

1. applying conversion hooks
2. runtime CA Bundle modification. 

I am open to suggestions if we have a generic approach that we can leverage but as of writing this message, 


> to conditionally enable webhooks if the type is registered within the scheme.

this can easily be done programatically already with existing resources and helper we provide right ?

As for the CA Bundle thing, if we have an infra such as Cert Manager running on the test cluster, they can do these changes/injection already. Do we really need to add /handle those in the e2e-framework?

klient/decoder/decoder.go
```diff
@@ -55,6 +56,18 @@ type MutateFunc func(obj k8s.Object) error
 // HandlerFunc is a function executed after an object has been decoded and patched. If an error is returned, further decoding is halted.
 type HandlerFunc func(ctx context.Context, obj k8s.Object) error
 
+// PreHandlerFuncHook is a handler helper provided to hook into the Workflow of HandlerFunc to make sure you can control what resources
+// are installed programatically in case if required. This can help in performing controlled CRUD operaton of the k8s resource using the
+// decoder package
+type PreHandlerFuncHook func(ctx context.Context, obj k8s.Object) bool
```
vladimirvivien, line 62: This seems to be adding yet another level of control when accessing the flow of decoding/encoding execution.  The `decoder` functions already provide ways of controlling what resources are selected via the glob patterns.  Can you give an example when you think this would be needed ?

klient/decoder/decoder.go
```diff
@@ -55,6 +56,18 @@ type MutateFunc func(obj k8s.Object) error
...
+
+// AllowAllHook is a simple allow All filter hook that is enabled by default
+func AllowAllHook() PreHandlerFuncHook {
```
vladimirvivien, line 65: See comment above.

klient/decoder/decoder.go
```diff
@@ -92,6 +105,30 @@ func DecodeAllFiles(ctx context.Context, fsys fs.FS, pattern string, options ...
 	return objects, err
 }
 
+// ApplyUsingManifestDir resolves all the files in the Directory dirPath against the globbing pattern and creates a kubernetes
+// resource for each of the resources found under the manifest directory.
+func ApplyUsingManifestDir(ctx context.Context, r *resources.Resources, dirPath, pattern string, f PreHandlerFuncHook, createOptions []resources.CreateOption, options ...DecodeOption) error {
+	err := DecodeEachFile(ctx, os.DirFS(dirPath), pattern, func(ctx context.Context, obj k8s.Object) error {
+		if f(ctx, obj) {
```
vladimirvivien, line 112: Again, I think this check should not be applied. We assume if someone call this function, it is because they want to apply the files. I don't the additional check is needed.

klient/decoder/decoder.go
```diff
@@ -92,6 +105,30 @@ func DecodeAllFiles(ctx context.Context, fsys fs.FS, pattern string, options ...
 	return objects, err
 }
 
+// ApplyUsingManifestDir resolves all the files in the Directory dirPath against the globbing pattern and creates a kubernetes
+// resource for each of the resources found under the manifest directory.
+func ApplyUsingManifestDir(ctx context.Context, r *resources.Resources, dirPath, pattern string, f PreHandlerFuncHook, createOptions []resources.CreateOption, options ...DecodeOption) error {
```
vladimirvivien, line 110: Suggestion: `ApplyWithManifestDir` to stay consistent with naming.

pkg/envfuncs/resource_funcs.go
```diff
@@ -0,0 +1,57 @@
...
+	"sigs.k8s.io/e2e-framework/klient/decoder"
+	"sigs.k8s.io/e2e-framework/klient/k8s/resources"
+	"sigs.k8s.io/e2e-framework/pkg/env"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+)
+
+// SetupCRDs is provided as a helper env.Func handler that can be used to setup the CRDs that are required
+// to process your controller code for testing. This env.Func helper also provides a helper function that can be
+// used to programatically filter selective resources under question as well all support injecting resources.CreateOption
+func SetupCRDs(crdPath, pattern string, opts []resources.CreateOption, f decoder.PreHandlerFuncHook, decodeOpts ...decoder.DecodeOption) env.Func {
```
vladimirvivien, line 31: I would make this function simpler, like:

```go
func SetupCRDs(crdPath, pattern string, opts []resources.CreateOption) env.Func { ... }
```
If someone needs more control, they can always call the `ApplyUsingManifestDir` which accepts more parameters.

vladimirvivien (commented):
@harshanarayana Thank you for updated code. However, I did not quite understand the need for the pre-hook hook. I dont think it's needed and it seems to make the code a bit more complex. Let me know why you think it's needed.

klient/decoder/decoder.go
harshanarayana: The idea behind this was to be able to provide a way for end users to programatically skip certain resource discovered.  
But then again the same can be done via the Pattern regex as well. Should I keep this or do you think it is better to cleanup this ?

Say I build the code once but I want to control what resource gets processed via the test, I can use this conditional check as a way to control that. But I might be over thinking this.

klient/decoder/decoder.go
harshanarayana: /ack

klient/decoder/decoder.go
harshanarayana: /ack

klient/decoder/decoder.go
harshanarayana: Done

pkg/envfuncs/resource_funcs.go
harshanarayana: I will cleanup the `f decoder.PreHandlerFuncHook` as I am cleaning it up entirely. The `decodeOpts` is still helpful to be able to use right ?

klient/decoder/decoder.go
vladimirvivien: @harshanarayana LoL.. I think you are overthinking this one. 
Remember, a test user can specify the name or name patterns for the resource file.
I just think that this is a bit overkill, specially  when I can specify the file name or a name pattern to select the resource file from a directory.

harshanarayana: :) I have cleaned it up.

pkg/envfuncs/resource_funcs.go
```diff
@@ -0,0 +1,51 @@
...
+	"sigs.k8s.io/e2e-framework/klient/decoder"
+	"sigs.k8s.io/e2e-framework/klient/k8s/resources"
+	"sigs.k8s.io/e2e-framework/pkg/env"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+)
+
+// SetupCRDs is provided as a helper env.Func handler that can be used to setup the CRDs that are required
+// to process your controller code for testing. This env.Func helper also provides support for
+// injecting resources.CreateOption
+func SetupCRDs(crdPath, pattern string, opts []resources.CreateOption, decodeOpts ...decoder.DecodeOption) env.Func {
```
vladimirvivien, line 31: I would keep `SetupCRDs` simple and remove the options, like:

```go
SetupCRDs(crdPath, pattern string) env.Func{...}
```
If someone needs those options, the person can use `decoder.ApplyWithManifestDir` directly to take control.

pkg/envfuncs/resource_funcs.go
```diff
@@ -0,0 +1,51 @@
...
+		if err != nil {
+			return ctx, err
+		}
+		return ctx, decoder.ApplyWithManifestDir(ctx, r, crdPath, pattern, opts, decodeOpts...)
+	}
+}
+
+// TeardownCRDs is provided as a handler function that can be hooked into your test's teardown sequence to
+// make sure that you can cleanup the CRDs that were setup as part of the SetupCRDs hook
+func TeardownCRDs(crdPath, pattern string, opts []resources.DeleteOption, decodeOpts ...decoder.DecodeOption) env.Func {
```
vladimirvivien, line 43: See comment above.

vladimirvivien (commented):
This looks good, left comments about method signature.

pkg/envfuncs/resource_funcs.go
harshanarayana: All right. Changes published as suggested

pkg/envfuncs/resource_funcs.go
harshanarayana: All right. Changes published as suggested

harshanarayana:
@vladimirvivien  PTAL. Suggestions have been addressed

vladimirvivien (approved):
LGTM.
Thanks again. This will be really helpful to support CRD-based frameworks such as CAPI, etc.

crandles:
great work here!

PR #143, opened 2022-06-05
GIT-141: enable panic handlers for ensuring Finish Steps

harshanarayana:


pkg/env/env.go
```diff
@@ -343,19 +343,30 @@ func (e *testEnv) Run(m *testing.M) int {
...
-	exitCode := m.Run() // exec test suite
-
-	finishes := e.getFinishActions()
-	// attempt to gracefully clean up.
-	// Upon error, log and continue.
-	for _, fin := range finishes {
-		// context passed down to each finish step
-		if e.ctx, err = fin.run(e.ctx, e.cfg); err != nil {
-			klog.V(2).ErrorS(err, "Finish action handlers")
+	defer func() {
```
harshanarayana, line 346: Which one do you think is the better place to handle panic? I added here as this was the only place where the `Finish` steps are being performed. But I am sort of conflicted. Should I move this to `processTests` instead ? cc @vladimirvivien  @ShwethaKumbla

ShwethaKumbla: As we are adding this feature to handle clean up/run finish task when panics this place looks good to me.

pkg/env/env.go
```diff
@@ -343,19 +343,30 @@ func (e *testEnv) Run(m *testing.M) int {
...
+		// Recover and see if the panic handler is disabled. If it is disabled, panic and stop the workflow.
+		// Otherwise, log and continue with running the Finish steps of the Test suite
+		rErr := recover()
+		if rErr != nil {
+			if e.cfg.IgnorePanicRecovery() {
+				panic(rErr)
+			}
+			klog.Error("Recovering from a panic", rErr)
```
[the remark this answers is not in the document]
harshanarayana: Ack. Let me make the changes accordingly.

vladimirvivien (commented):
Good start. I think you are almost there.

pkg/env/env.go
harshanarayana: Ack.

harshanarayana:
@vladimirvivien  @ShwethaKumbla  I've addressed the changes request by both of you. PTAL when time permits. Thanks

pkg/env/env.go
```diff
@@ -336,26 +336,38 @@ func (e *testEnv) Run(m *testing.M) int {
...
+			if e.cfg.IgnorePanicRecovery() {
+				panic(rErr)
+			}
+			klog.Error("Recovering from panic and running finish actions", rErr)
+		}
+
+		finishes := e.getFinishActions()
+		// attempt to gracefully clean up.
+		// Upon error, log and continue.
+		for _, fin := range finishes {
```
vladimirvivien, line 354: What happens if panic occurs in Finish steps ?
Will it recover  again ?
If so, should we check to see if panic happened during Finish step ?

pkg/flags/flags.go
```diff
@@ -26,17 +26,18 @@ import (
...
+	flagFeatureName         = "feature"
+	flagAssessName          = "assess"
+	flagLabelsName          = "labels"
+	flagSkipLabelName       = "skip-labels"
+	flagSkipFeatureName     = "skip-features"
+	flagSkipAssessmentName  = "skip-assessment"
+	flagParallelTestsName   = "parallel"
+	flagDryRunName          = "dry-run"
+	flagFailFast            = "fail-fast"
+	flagIgnorePanicRecovery = "ignore-panic-recovery"
```
vladimirvivien, line 40: I struggle with the name of the flag.
I think we should name the flag as it relates to the features of the framework, not as a function of how Go works.
What about `--disable-graceful-teardown` or `--force-finish`. That way it relates to how the framework functions.

vladimirvivien (commented):
Great progress. I left comment about the name of the flag itself.

pkg/flags/flags.go
harshanarayana: No worries. I do like the `--disable-graceful-teardown`. I will update it to this one.

vladimirvivien (approved):
I like the fixes, thank you.

/lgtm

PR #145, opened 2022-06-07
docs(readme): fix test example

ernado:
Fix compiling errors in test function example from readme.

* nspaces should be `corev1.NamespaceList` type
* testenv.Test signature is `Test(*testing.T, ...Feature)`, so t should be first arg


vladimirvivien (approved):
/lgtm

PR #147, opened 2022-06-09
Update example for failfast commandline flag

ShwethaKumbla:
This PR is for updating the example shown in README file in to go file.

### DIFF
--- a/examples/fail_fast/fail_fast_test.go
+++ b/examples/fail_fast/fail_fast_test.go
@@ -1,0 +1,72 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package fail_fast
+
+import (
+	"context"
+	"testing"
+
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+func TestExample(t *testing.T) {
+	failFeature := features.New("fail-feature").
+		Assess("1==2", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			if 1 != 2 {
+				t.Log("1 != 2")
+				t.FailNow() // mark test case as failed here, don't continue execution
+			} else {
+				t.Log("1 == 2")
+			}
+			return ctx
+		}).
+		Assess("print", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			t.Log("THIS LINE SHOULDN'T BE PRINTED")
+			return ctx
+		}).
+		Teardown(func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+			t.Log("This teardown should not be invoked")
+			return ctx
+		}).
+ 		Feature()
+
+	nextFeature := features.New("next-feature").
+		Assess("print", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			t.Log("THIS LINE ALSO SHOULDN'T BE PRINTED")
+			return ctx
+		}).
+		Feature()
+
+	testenv.Test(t, failFeature, nextFeature)
+}
+
+// even if the previous testcase fails, execute this testcase
+func TestNext(t *testing.T) {
+	nextFeature := features.New("next-test-feature").
+		Assess("print", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			t.Log("THIS LINE SHOULD BE PRINTED")
+			return ctx
+		}).
+		Teardown(func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+			t.Log("This teardown should be invoked")
+			return ctx
+		}).
+		Feature()
+
+	testenv.Test(t, nextFeature)
+}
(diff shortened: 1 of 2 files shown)


harshanarayana (changes requested):
This test was kept only in document's README because adding this will cause the tests to fail as this is an error case. :( Not sure if there is a clear way to add this test in a place other than README. Or we should run this with a build tag and ignore that during run. So we will have a file but it won't run

ShwethaKumbla:
Hmm one of the user was having issue while using `--fail-fast ` flag so I quickly tried this example to show them how to use. And raised a PR :) I dint realise that it's an error scenario and the test job would fail.

Let me check is there any way we could include such tests.

harshanarayana:
> Hmm one of the user was having issue while using `--fail-fast ` flag so I quickly tried this example to show them how to use. And raised a PR :) I dint realise that it's an error scenario and the test job would fail.
> 
> Let me check is there any way we could include such tests.

@ShwethaKumbla was there an error with the behavior of the flag?

harshanarayana:
One way I can think of is use a `// +build exampleonly` so that regular tests won't run them but we can still have it in place without breaking builds

ShwethaKumbla:
@harshanarayana Its a known behaviour. Without t.Fail or t.FailNow framework is not recognising the flag.
image

harshanarayana:
@ShwethaKumbla  No that should not be the case, May be you are using a wrong version of the `e2e-framework` ?

```bash
❯ ls                                                                                                                                                                                                                                                                 77%  ▓▒░
README.md         fail.test         fail_fast_test.go main_test.go

```
```bash
❯ go test . -test.v -args --fail-fast                                                                                                                                                                                                                                77%  ▓▒░
=== RUN   TestExample
=== RUN   TestExample/fail-feature
=== RUN   TestExample/fail-feature/1==2
    fail_fast_test.go:31: 1 != 2
--- FAIL: TestExample (0.00s)
    --- FAIL: TestExample/fail-feature (0.00s)
        --- FAIL: TestExample/fail-feature/1==2 (0.00s)
=== RUN   TestNext
=== RUN   TestNext/next-test-feature
=== RUN   TestNext/next-test-feature/print
    fail_fast_test.go:62: THIS LINE SHOULD BE PRINTED
=== CONT  TestNext/next-test-feature
    fail_fast_test.go:66: This teardown should be invoked
--- PASS: TestNext (0.00s)
    --- PASS: TestNext/next-test-feature (0.00s)
        --- PASS: TestNext/next-test-feature/print (0.00s)
FAIL
FAIL	sigs.k8s.io/e2e-framework/examples/fail_fast	0.419s
FAIL
```

harshanarayana:
```bash
❯ ./fail.test --help 2>&1 | grep -A1 fail-fast                                                                                                                                                                                                                       77%  ▓▒░
  -fail-fast
    	Fail immediately and stop running untested code

**Comment by ShwethaKumbla** [comment]
@harshanarayana I am using the latest version. The problem is under examples/fail_fast folder its working fine. If you try out `-args --fail-fast` under other test folders its giving error saying that flag is not defined.  I am seeing this behaviour so could you please check from your side even you see the same?

sk2@sk2-a01 decoder % 
```
decoder % ls                      
README.md	decoder_test.go	main_test.go	testdata
decoder % go test . -args --fail-fast
flag provided but not defined: -fail-fast
Usage of /var/folders/gt/5n_dcxq51wl5mz26lbgy2c6m0000gp/T/go-build900183505/b001/decoder.test:
  -test.bench regexp
    	run only benchmarks matching regexp
  -test.benchmem
    	print memory allocations for benchmarks
  -test.benchtime d
    	run each benchmark for duration d (default 1s)
  -test.blockprofile file
    	write a goroutine blocking profile to file
  -test.blockprofilerate rate
    	set blocking profile rate (see runtime.SetBlockProfileRate) (default 1)
  -test.count n
    	run tests and benchmarks n times (default 1)
  -test.coverprofile file
    	write a coverage profile to file
  -test.cpu list
    	comma-separated list of cpu counts to run each test with
  -test.cpuprofile file
    	write a cpu profile to file
  -test.failfast
    	do not start new tests after the first test failure
  -test.list regexp
    	list tests, examples, and benchmarks matching regexp then exit
  -test.memprofile file
    	write an allocation profile to file
  -test.memprofilerate rate
    	set memory allocation profiling rate (see runtime.MemProfileRate)
  -test.mutexprofile string
    	write a mutex contention profile to the named file after execution
  -test.mutexprofilefraction int
    	if >= 0, calls runtime.SetMutexProfileFraction() (default 1)
  -test.outputdir dir
    	write profiles to dir
  -test.paniconexit0
    	panic on call to os.Exit(0)
  -test.parallel n
    	run at most n tests in parallel (default 16)
  -test.run regexp
    	run only tests and examples matching regexp
  -test.short
    	run smaller test suite to save time
  -test.shuffle string
    	randomize the execution order of tests and benchmarks (default "off")
  -test.testlogfile file
    	write test action log to file (for use only by cmd/go)
  -test.timeout d
    	panic test binary after duration d (default 0, timeout disabled)
  -test.trace file
    	write an execution trace to file
  -test.v
    	verbose: print additional output
FAIL	sigs.k8s.io/e2e-framework/examples/decoder	43.215s
FAIL
decoder % 
```
Probably we should document the usage for fail-fast flag if above is the case.

**Comment by harshanarayana** [comment]
> The problem is under examples/fail_fast folder its working fine. If you try out `-args --fail-fast` under other test folders its giving error saying that flag is not defined.

That is because it only sets up the flag if you use the `cfg, _ := envconf.NewFromFlags()` to setup the flags under pflags. Not all of our examples use that. 

```bash
cfg, _ := envconf.NewFromFlags()
testenv = env.NewWithConfig(cfg)
```

vs

```bash
testenv = env.New()
```

ShwethaKumbla:
Oh yes, we have two ways to set the env var. My bad , totally forgot about it. 🤦‍♀️

ShwethaKumbla:
Adding this example is leading to job failure as this example is to demonstrate how to fail fast the tests. 
Its been already captured in README.md file under examples/fail_fast and we will look back again in future to handle such examples as part of PR jobs. Hence closing this PR.

PR #156, opened 2022-07-08
controller-runtime: enable ability to fetch controller runtime client

### LINKED ISSUE #146: Enhancement: Able to get controller-runtime Client from the Resources
In most of the cases, we need to get and utilize controller-runtime Client internally for the intermediate functions
It would be nice to have a functions GetClient() inside the `klient/k8s/resources/resources.go` to get controller-runtime Client for this case.
Alternatively, we could probably have a helper function in the `klient` package returning Client as
```golang
cr.New(c.cfg, cr.Options{Scheme: scheme.Scheme})
```

harshanarayana:
## What does this do ?
Enable the ability to fetch the reusable controller-runtime client via the `e2e-framework` infra. 

## Why is this required?
https://github.com/kubernetes-sigs/e2e-framework/issues/146#issuecomment-1177699723 Has a good explanation on the usability of this features. 

Fixes #146

### DIFF
--- a/klient/client.go
+++ b/klient/client.go
@@ -17,7 +17,9 @@ limitations under the License.
 package klient
 
 import (
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
+	cr "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/e2e-framework/klient/conf"
 	"sigs.k8s.io/e2e-framework/klient/k8s/resources"
 )
@@ -36,6 +38,12 @@ type Client interface {
 type client struct {
 	cfg       *rest.Config
 	resources *resources.Resources
+}
+
+// NewControllerRuntimeClient provides an instance of the Controller runtime client with
+// the provided rest config and custom runtime scheme.
+func NewControllerRuntimeClient(cfg *rest.Config, scheme *runtime.Scheme) (cr.Client, error) {
+	return cr.New(cfg, cr.Options{Scheme: scheme})
 }
 
 // New returns a new Client value
--- a/klient/k8s/resources/resources.go
+++ b/klient/k8s/resources/resources.go
@@ -189,6 +189,11 @@ func (r *Resources) GetScheme() *runtime.Scheme {
 	return r.scheme
 }
 
+// GetClient return the controller-runtime client instance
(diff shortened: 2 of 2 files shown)


klient/k8s/resources/resources.go
```diff
@@ -189,6 +189,11 @@ func (r *Resources) GetScheme() *runtime.Scheme {
 	return r.scheme
 }
 
+// GetClient return the controller-runtime client instance
+func (r *Resources) GetClient() cr.Client {
```
vladimirvivien, line 193: Because there is function `klient.New()` that returns a `klient.Client`. Maybe this should be named to reflect what it is returning i.e., `GetRawClient` or `GetCRClient` or `GetControllerRuntimeClient`

klient/client.go
```diff
@@ -38,6 +40,12 @@ type client struct {
 	resources *resources.Resources
 }
 
+// NewControllerRuntimeClient provides an instance of the Controller runtime client with
+// the provided rest config and custom runtime scheme.
+func NewControllerRuntimeClient(cfg *rest.Config, scheme *runtime.Scheme) (cr.Client, error) {
```
vladimirvivien, line 45: I am not sure if this is needed.  What I do think is needed is a way to inject an existing Controller Runtime client to be used be used with Resources:

```go
func (r *Resources) SetCRClient(c cr.Client) {
  r.client = c
}
```
Or something like that

vladimirvivien (commented):
Thanks for the prompt PR.
I left couple of comments.

klient/client.go
harshanarayana: @vladimirvivien  THis might break a few things if we are not careful possibly ?

```golang
type Resources struct {
	config *rest.Config
	scheme *runtime.Scheme
	client cr.Client
	namespace string
}
```

If we are to override the `Resources.client` should we allow the same for `Resources.config` if so, how do we enforce that? Document it and hope everyone will read and keep that in mind ? Might get a bit messy is it not ?

harshanarayana: I think we should not let someone set the `Resources.Client` directly. Instead, use the `klient/client.go` method to create a new resource for custom config or use the `GetControllerRuntimeClient` function from the `Resources` to access the already existing client. That way, we can keep this clean.

vladimirvivien (approved):
This looks good to me. Thanks @harshanarayana

Issue #161, opened 2022-09-07
Wrong log output when failing during BeforeEachFeature, AfterEachFeature & AfterEachTest


mhofstetter:
**Current log output**

Errors during test / feature setup & cleanup functions are always reported as an error during`BeforeEachTest`.

* Error during `BeforeEachTest` -> `BeforeEachTest failure: xxx`
* Error during `BeforeEachFeature` -> `BeforeEachTest failure: xxx`
* Error during `AfterEachFeature` -> `BeforeEachTest failure: xxx`
* Error during `AfterEachTest` -> `BeforeEachTest failure: xxx`

e.g.

```go
testenv.AfterEachFeature(func(ctx context.Context, config *envconf.Config, t *testing.T, f features.Feature) (context.Context, error) {
	return ctx, errors.New("error in AfterEachFeature")
})
```

results in 

```sh
=== CONT  Test
    env.go:217: BeforeEachTest failure: error in AfterEachFeature
```
**Expected log output**

Error during before/after actions should be reported with the correct "action role".

* Error during `BeforeEachTest` -> `BeforeEachTest failure: xxx`
* Error during `BeforeEachFeature` -> `BeforeEachFeature failure: xxx`
* Error during `AfterEachFeature` -> `AfterEachFeature failure: xxx`
* Error during `AfterEachTest` -> `AfterEachTest failure: xxx`

harshanarayana:
@mhofstetter Thanks for this report. This was a mistake that got introduced in one of my old changes. Thanks for the PR to correct this.

PR #162, opened 2022-09-07
Fix log output on errors during test lifecycle actions

mhofstetter:
Hi

This change should fix https://github.com/kubernetes-sigs/e2e-framework/issues/161 by logging the correct action role as prefix when failing during a test lifecycle action.

Thanks for having a look in advance & keep up the good work!


vladimirvivien (commented):
Thank you for your changes. They look good.

vladimirvivien (approved):

Issue #194, opened 2023-01-26
How to specify multiple labels when running tests


embano1:
This example recommends a label key `test` and values for individual features:

https://github.com/kubernetes-sigs/e2e-framework/blob/8e4361f050f28ea6dc55e8885a14f7d66dd4d45a/README.md?plain=1#L70-L84

However, following this logic, I cannot run multiple features with the same label `key`, e.g. 

`--labels="feature=feat1,feature=feat2"` is not possible (in my current tests the first one is ignored). I guess this is due to the label parsing map used:

https://github.com/kubernetes-sigs/e2e-framework/blob/8e4361f050f28ea6dc55e8885a14f7d66dd4d45a/pkg/flags/flags.go#L290

Wondering if we should fix this so I can run multiple features instead of just one when using the same label `key`.

vladimirvivien:
@embano1 thanks for the issue. Are you saying that a features cannot have multiple labels added to it? Or are you saying you just want to run multiple features with the same labels ? 

For the latter, you should be able to declare multiple features with the same labels and run them:

```go
func TestHello_WithSetup(t *testing.T) {
	e := env.NewWithConfig(envconf.New())
	var name string
	f0 := features.New("Hello Feature").
		WithLabel("type", "hello").
		Assess("test message", func(ctx context.Context, t *testing.T, _ *envconf.Config) context.Context {
			result := Hello(name)
			if result != "Hello foobar" {
				t.Error("unexpected message")
			}
			return ctx
		}).Feature()

	  f1 := features.New("Goodbye Feature").
		WithLabel("type", "message").
		Assess("test message", func(ctx context.Context, t *testing.T, _ *envconf.Config) context.Context {
			result := Goodbye(name)
			if result != "Goodbye foobar" {
				t.Error("unexpected message")
			}
			return ctx
		}).Feature()
	e.Test(t, f0, f1)
}
```
You should be able to select each feature by label as follows:

```
go run . -args --labels="type=hello, type=goodbye"
```

If i completely misunderstood your question, please let me know.

embano1:
Thx for your response! What I found was that if `featureX` had label `"feature=foo"` and `featureY` had the label `"feature=bar` if I run `go run . -args --labels="feature=foo,feature=bar"` only `featureY` ran. Looking up the code, it seems a map is used which would overwrite labels with the same key i.e. `"feature`.

PR #196, opened 2023-01-31
fix: Parse multiple label values for same key

embano1:
Changes `--labels` and `--skip-labels` parsing logic to allow for multiple values for the same label `key`.

Fixes: #194

### DIFF
--- a/pkg/features/builder.go
+++ b/pkg/features/builder.go
@@ -34,7 +34,7 @@ func New(name string) *FeatureBuilder {
 
 // WithLabel adds a test label key/value pair
 func (b *FeatureBuilder) WithLabel(key, value string) *FeatureBuilder {
-	b.feat.labels[key] = value
+	b.feat.labels[key] = append(b.feat.labels[key], value)
 	return b
 }
 
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -280,10 +280,10 @@ func ParseArgs(args []string) (*EnvFlags, error) {
 	}, nil
 }
 
-type LabelsMap map[string]string
+type LabelsMap map[string][]string
 
 func (m LabelsMap) String() string {
-	i := map[string]string(m)
+	i := map[string][]string(m)
 	return fmt.Sprint(i)
 }
 
@@ -295,8 +295,19 @@ func (m LabelsMap) Set(val string) error {
 		if len(kv) != 2 {
 			return fmt.Errorf("label format error: %s", label)
 		}
-		m[strings.TrimSpace(kv[0])] = strings.TrimSpace(kv[1])
+		k := strings.TrimSpace(kv[0])
+		v := strings.TrimSpace(kv[1])
+		m[k] = append(m[k], v)
 	}
 
 	return nil
+}
+
+func (m LabelsMap) Contains(key, val string) bool {
+	for _, v := range m[key] {
+		if val == v {
+			return true
+		}
+	}
+	return false
 }
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -512,20 +512,24 @@ func (e *testEnv) requireProcessing(kind, testName string, requiredRegexp, skipR
 	}
 
 	if labels != nil {
-		for k, v := range e.cfg.Labels() {
-			if labels[k] != v {
-				skip = true
-				message = fmt.Sprintf(`Skipping feature "%s": unmatched label "%s=%s"`, testName, k, labels[k])
-				return skip, message
+		for key, vals := range e.cfg.Labels() {
+			for _, v := range vals {
+				if !labels.Contains(key, v) {
+					skip = true
+					message = fmt.Sprintf(`Skipping feature "%s": unmatched label "%s=%s"`, testName, key, v)
+					return skip, message
+				}
 			}
 		}
 
 		// skip running a feature if labels matches with --skip-labels
-		for k, v := range e.cfg.SkipLabels() {
-			if labels[k] == v {
-				skip = true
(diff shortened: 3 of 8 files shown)


pkg/flags/flags.go
```diff
@@ -295,8 +295,19 @@ func (m LabelsMap) Set(val string) error {
 		if len(kv) != 2 {
 			return fmt.Errorf("label format error: %s", label)
 		}
-		m[strings.TrimSpace(kv[0])] = strings.TrimSpace(kv[1])
+		k := strings.TrimSpace(kv[0])
```
vladimirvivien, line 298: Probably would be a good idea to save both key and values all lowercases.

vladimirvivien (commented):
@embano1 thank you for the awesome contribution. At a high-level, this looks good to me. Pinging @harshanarayana for a second opinion.

pkg/flags/flags.go
embano1: Yup, good idea!

embano1:
> @embano1 thank you for the awesome contribution. At a high-level, this looks good to me. Pinging @harshanarayana for a second opinion.

Thx! There was a bit of refactoring involved, so please let me know if you're all happy with that. No prob to change...

pkg/flags/flags.go
vladimirvivien: @embano1 Maybe just the key should be lowered and be case insensitive so that it makes searching for a label simpler (in the `Contains` method) or as an alternative use `strings.EqualFold` when comparing in `Contains`.

harshanarayana:
Let me check this first thing tomorrow.

pkg/flags/flags.go
embano1: Made `key` case-insensitive in `Set()` and `Contains()` and updated `flag` doc string.

harshanarayana:
Thanks @embano1  for this great contribution. Looks good for me over all. Marking it ok for tests to be run. 

/ok-to-test

docs/design/test-harness-framework.md
```diff
@@ -462,7 +462,8 @@ The framework shall provide predefined flagsets that will be automatically appli
 
 * `--feature` - a regular expression that target features to run
 * `--assess` - a regular expression that targets the name of an assessment steps
-* `--labels` - a comma-separated list of key/value pairs  used to filter features by their assigned labels
+* `--labels` - a comma-separated list of key/value pairs  used to filter features by their assigned labels (key is
```
harshanarayana, line 465: Need to update the docs bit on `--skip-labels` as well. 

https://github.com/kubernetes-sigs/e2e-framework/blob/cd0fd6dcf181539c450224b217ac3c20f7cac185/docs/design/test-harness-framework.md?plain=1#L475

pkg/features/builder.go
```diff
@@ -34,7 +34,7 @@ func New(name string) *FeatureBuilder {
 
 // WithLabel adds a test label key/value pair
 func (b *FeatureBuilder) WithLabel(key, value string) *FeatureBuilder {
-	b.feat.labels[key] = value
+	b.feat.labels[key] = append(b.feat.labels[key], value)
```
harshanarayana, line 37: Since we process the argument keys on label related attribute as lower case, would it make sense to turn the key here as well into lowercase and store it in a uniform way ?

third_party/kubetest2/pkg/tester/e2e-tester.go
```diff
@@ -36,7 +36,7 @@ type Tester struct {
 	Kubeconfig              string `desc:"Path to a cluster kubeconfig file (optional)"`
 	Feature                 string `desc:"Regular expression to select feature(s) to test"`
 	Assess                  string `desc:"Regular expression to select assessment(s) to run"`
-	Labels                  string `desc:"Comma-separated key=value to filter features by labels"`
+	Labels                  string `desc:"Comma-separated key=value to filter features by labels (key is case-insensitive)"`
```
harshanarayana, line 39: Needs to update the description of `SkipLabels` as well.

harshanarayana (changes requested):
I might be nit picking but should we make the ignore case match an optional thing ? I hope no one is relying on the case sensitive matches to tag test cases labels differently, but if they do, this change can shadow labels. Or at least call it out in the release indicating that this will be a change that one has to keep an eye on ? cc @vladimirvivien  what do you think ?

@embano1  Great work. I just have a few minor suggestions that I am leaving on the PR to be addressed.

embano1:
@harshanarayana thx for your critical eye! How about reverting the case-insensitivity for now?

vladimirvivien:
I like the reviews. I am don't feel strongly (either way) about the case sensitivity. I know @embano1 has been going back and forth due to my suggestion. Nevertheless, if the final state includes case-insensitivity then we simply could just document it. If we revert back to sensitive cases (which i think that's how it worked originally) then nothing changes.  I think this is really close to getting in.

harshanarayana:
>  I think this is really close to getting in.

@vladimirvivien @embano1 Let us move it back to the original format for now (case sensitive) if @embano1  doesn't mind doing that and open a different issue to track the changes required for getting the case insensitive keys for label filters ?

Other than this one thing I don;t have any change requests from my end either.

embano1:
Yup, I'm also in favor of reverting and I'll create a tracking issue for the case-insensitivity.

pkg/flags/flags.go
embano1: Reverted as per the PR discussion

embano1:
Created a tracking issue #198 and added a test which will break once we add case-insensitive handling so we capture such changes in CI.

Issue #187, opened 2023-01-09
Add `--context` flag


maruina:
It would be great to be able to pass a `--context` flag to protect against running the tests in the wrong cluster.

I can see this useful when you might switch context in a different shell and then re-running the tests or to make sure to run the tests against a specific cluster.

I would be happy to raise the PR if you think this feature might be useful.

vladimirvivien:
It may be worth looking into exposing Kubernetes flags from the CLI-runtime project to handle most common cluster flags (https://github.com/kubernetes/cli-runtime).

vladimirvivien:
@maruina no worries, if you think you may want to take on that as a PR, I am here to help.

harshanarayana:
This would be a good addition to have. But we might have to think of a few cases

1. How will this fit in along with the `--kubeconfig` argument
2. If one were to setup unique  `kubeconfig` for each of their test run + cluster combination, the context is not really required as the `--kubeconfig` should be able to handle the requirement
3. How does this work when the new cluster is being created with the `kind` based setup ? The context passed here will be the cluster context for the newly created kind cluster ?
4. Do we have to now consider checking the actual context is valid or not before using it ?

maruina:
Hey @harshanarayana, thanks for your input. I think we can:
1. Follow the kubectl behaviour, where you can pass both flags, none or one or the other
2. Yes, you are correct. If you have a single context in the kubeconfig then you don't need this. In my case I have multiple contexts and when running the test I want to make sure I'm targeting the desired cluster
3. If you're not passing a context I would assume you use the current context, so there should be no changes for when you create a kind cluster
4. When you use the wrong context in kubectl you get
```
❯ kubectl --context foo get pods
Error in configuration: context was not found for specified context: foo
```
so I would just error when we try to use it. I don't know if we can make any checks when we use `clientcmd`

What do you think?

PR #197, opened 2023-02-02
Add --context flag

maruina:
Fix https://github.com/kubernetes-sigs/e2e-framework/issues/187

@vladimirvivien I know this is not what you suggested, but I spent some time looking at the cli-runtime and couldn't figure out how to wire it.
Since this is blocking our adoption and it is a pretty simple change, I'd like to ask you if we can still merge it and then I can try to take again a look at the cli-runtime fix

### DIFF
--- a/examples/flags/README.md
+++ b/examples/flags/README.md
@@ -68,3 +68,9 @@ To get additional verbose logs
 ```shell
 ./flags.test --assess es --v 2
 ```
+
+To run a test against a particular Kubeconfig context
+
+```shell
+./flags.test --kubeconfig ~/path/to/kubeconfig --context my-context
+```
--- a/klient/conf/config.go
+++ b/klient/conf/config.go
@@ -18,6 +18,7 @@ limitations under the License.
 package conf
 
 import (
+	"errors"
 	"flag"
 	"os"
 	"os/user"
@@ -33,14 +34,32 @@ var DefaultClusterContext = ""
 // New returns Kubernetes configuration value of type *rest.Config.
 // filename is kubeconfig file
 func New(fileName string) (*rest.Config, error) {
-	// if filename is not provided assume in-cluster-config
+	var resolvedKubeConfigFile string
+	kubeContext := ResolveClusterContext()
+
+	// resolve the kubeconfig file
+	resolvedKubeConfigFile = fileName
 	if fileName == "" {
-		return rest.InClusterConfig()
+		resolvedKubeConfigFile = ResolveKubeConfigFile()
 	}
 
-	// create the config object from k8s config path
+	// if resolvedKubeConfigFile is still empty, assume in-cluster config
+	if resolvedKubeConfigFile == "" {
+		if kubeContext == "" {
+			return rest.InClusterConfig()
+		}
+		// if in-cluster can't use the --kubeContext flag
+		return nil, errors.New("cannot use a cluster context without a valid kubeconfig file")
+	}
+
+	// set the desired context if provided
+	if kubeContext != "" {
+		return NewWithContextName(resolvedKubeConfigFile, kubeContext)
+	}
+
+	// create the config object from resolvedKubeConfigFile without a context
 	return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
-		&clientcmd.ClientConfigLoadingRules{ExplicitPath: fileName}, &clientcmd.ConfigOverrides{}).ClientConfig()
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: resolvedKubeConfigFile}, &clientcmd.ConfigOverrides{}).ClientConfig()
 }
 
 // NewWithContextName returns k8s config value of type *rest.Config
@@ -110,9 +129,9 @@ func ResolveKubeConfigFile() string {
 
 // ResolveClusterContext returns cluster context name based on --context flag.
 func ResolveClusterContext() string {
-	// If a flag --kube-context is specified use that
+	// If a flag --context is specified use that
 	if flag.Parsed() {
-		f := flag.Lookup("kube-context")
+		f := flag.Lookup("context")
 		if f != nil {
 			return f.Value.String()
 		}
--- a/pkg/envconf/config.go
+++ b/pkg/envconf/config.go
@@ -44,6 +44,7 @@ type Config struct {
 	dryRun                  bool
 	failFast                bool
 	disableGracefulTeardown bool
+	kubeContext             string
 }
 
 // New creates and initializes an empty environment configuration
@@ -85,6 +86,7 @@ func NewFromFlags() (*Config, error) {
 	e.dryRun = envFlags.DryRun()
 	e.failFast = envFlags.FailFast()
 	e.disableGracefulTeardown = envFlags.DisableGracefulTeardown()
+	e.kubeContext = envFlags.KubeContext()
 
 	return e, nil
 }
@@ -272,6 +274,17 @@ func (c *Config) WithDisableGracefulTeardown() *Config {
 // DisableGracefulTeardown is used to check the panic recovery handler should be enabled
 func (c *Config) DisableGracefulTeardown() bool {
 	return c.disableGracefulTeardown
+}
+
+// WithKubeContext is used to set the kubeconfig context
+func (c *Config) WithKubeContext(kubeContext string) *Config {
+	c.kubeContext = kubeContext
+	return c
+}
+
+// WithKubeContext is used to get the kubeconfig context
+func (c *Config) KubeContext() string {
+	return c.kubeContext
 }
 
 func randNS() string {
--- a/pkg/flags/flags.go
+++ b/pkg/flags/flags.go
@@ -38,6 +38,7 @@ const (
 	flagDryRunName              = "dry-run"
 	flagFailFast                = "fail-fast"
 	flagDisableGracefulTeardown = "disable-graceful-teardown"
+	flagContext                 = "context"
 )
 
 // Supported flag definitions
@@ -90,6 +91,10 @@ var (
 		Name:  flagDisableGracefulTeardown,
 		Usage: "Ignore panic recovery while running tests. This will prevent test finish steps from getting executed on panic",
 	}
+	contextFlag = flag.Flag{
+		Name:  flagContext,
+		Usage: "The name of the kubeconfig context to use",
+	}
 )
 
 // EnvFlags surfaces all resolved flag values for the testing framework
@@ -106,6 +111,7 @@ type EnvFlags struct {
 	dryRun                  bool
 	failFast                bool
 	disableGracefulTeardown bool
+	kubeContext             string
 }
 
 // Feature returns value for `-feature` flag
@@ -180,6 +186,11 @@ func Parse() (*EnvFlags, error) {
 	return ParseArgs(os.Args[1:])
 }
 
+// Context returns an optional kubeconfig context to use
+func (f *EnvFlags) KubeContext() string {
+	return f.kubeContext
+}
+
 // ParseArgs parses the specified args from global flag.CommandLine
 // and returns a set of environment flag values.
 func ParseArgs(args []string) (*EnvFlags, error) {
@@ -194,6 +205,7 @@ func ParseArgs(args []string) (*EnvFlags, error) {
 		dryRun                  bool
 		failFast                bool
 		disableGracefulTeardown bool
+		kubeContext             string
 	)
 
 	labels := make(LabelsMap)
@@ -247,6 +259,10 @@ func ParseArgs(args []string) (*EnvFlags, error) {
 		flag.BoolVar(&disableGracefulTeardown, disableGracefulTeardownFlag.Name, false, disableGracefulTeardownFlag.Usage)
 	}
 
+	if flag.Lookup(contextFlag.Name) == nil {
+		flag.StringVar(&kubeContext, contextFlag.Name, contextFlag.DefValue, contextFlag.Usage)
+	}
+
 	// Enable klog/v2 flag integration
 	klog.InitFlags(nil)
 
@@ -277,6 +293,7 @@ func ParseArgs(args []string) (*EnvFlags, error) {
 		dryRun:                  dryRun,
 		failFast:                failFast,
 		disableGracefulTeardown: disableGracefulTeardown,
+		kubeContext:             kubeContext,
 	}, nil
 }
 
--- a/klient/conf/main_test.go
+++ b/klient/conf/main_test.go
@@ -40,6 +40,7 @@ func setup() {
 
 	kubeconfigdir := filepath.Join(home, "test", ".kube")
 	kubeconfigpath := filepath.Join(kubeconfigdir, "config")
+	kubeContext := "test-context"
 
 	// check if file exists
 	_, err := os.Stat(kubeconfigpath)
@@ -52,7 +53,7 @@ func setup() {
 		}
 
 		// generate kube config data
-		data := genKubeconfig("test-context")
+		data := genKubeconfig(kubeContext)
 
 		err = createFile(kubeconfigpath, data)
 		if err != nil {
@@ -64,11 +65,19 @@ func setup() {
 	log.Info("file created successfully", kubeconfigpath)
 
 	flag.StringVar(&kubeconfig, "kubeconfig", "", "Paths to a kubeconfig. Only required if out-of-cluster.")
+	flag.StringVar(&kubeContext, "context", "", "The name of the kubeconfig context to use. Only required if out-of-cluster.")
 
 	// set --kubeconfig flag
 	err = flag.Set("kubeconfig", kubeconfigpath)
 	if err != nil {
-		log.ErrorS(err, "unexpected error while setting flag value")
+		log.ErrorS(err, "unexpected error while setting kubeconfig flag value")
+		return
+	}
+
+	// set --context flag
+	err = flag.Set("context", kubeContext)
+	if err != nil {
+		log.ErrorS(err, "unexpected error while setting context flag value")
 		return
 	}
 


pkg/envconf/config.go
```diff
@@ -44,6 +44,7 @@ type Config struct {
 	dryRun                  bool
 	failFast                bool
 	disableGracefulTeardown bool
+	context                 string
```
vladimirvivien, line 47: To avoid confusion with Go context, I would suggest naming this `k8sContext`

pkg/flags/flags.go
```diff
@@ -175,6 +181,11 @@ func (f *EnvFlags) DisableGracefulTeardown() bool {
 	return f.disableGracefulTeardown
 }
 
+// Context returns an optional kubeconfig context to use
+func (f *EnvFlags) Context() string {
```
vladimirvivien, line 185: Same here, I would suggest naming the method `KubernetesContext`

vladimirvivien (commented):
@maruina Hi. Thanks for the contribution. This is a good start.
I left some comments about the `context` name. Also, I don't see where in the code you are passing the kubernetes context to the the client.  I imagine that is what you want, correct ?

pkg/envconf/config.go
maruina: We have a `kubeconfig`, what do you think about `kubecontext`? I am asking so we have only one way to prefix kubernetes things

maruina:
> Also, I don't see where in the code you are passing the kubernetes context to the the client. I imagine that is what you want, correct ?

No :P
You are right and having a better look at the code, correct me where I'm wrong:
- we build the client with [ NewClient()](https://github.com/kubernetes-sigs/e2e-framework/blob/33c3d5efc189c44ff7dbc283a361481a264c8005/pkg/envconf/config.go#L111) where we pass the kubeconfig we got from `NewFromFlags`
- Then we call [`NewWithKubeConfigFile`](https://github.com/kubernetes-sigs/e2e-framework/blob/33c3d5efc189c44ff7dbc283a361481a264c8005/klient/client.go#L59) which is calling `conf.New(filePath)`
- If I look at the `klient/conf/config.go` I see we already have a [NewWithContextName](https://github.com/kubernetes-sigs/e2e-framework/blob/33c3d5efc189c44ff7dbc283a361481a264c8005/klient/conf/config.go#L47), but it seems we don't use it anywhere

What I could do is:
- check if`c.kubecontext` is empty or not in `NewClient()`
- call something like `NewWithKubeConfigFileAndContext()` which then calls `NewWithContextName`

It sounds a little convoluted so I'm not 100% how would you like to do it or if you have a better suggestion

vladimirvivien:
@maruina oh yes, you are right. There is already a method that can apply the Kubernetes cluster context at test time.

So, my next question to you, do you have to pass your cluster context name as a CLI argument ? You could use an environment variable (i.e. KUBE_CONTEXT) and read it from your test, then apply that method. That way you won't even have to wait for PR or a code change.

vladimirvivien:
@maruina I went and look at the code again. If you definitely want to move with this, introduce a `--context` as a flag, there are some additional steps you will need for it to work. I think you are almost there, but here is the complete path of changes you need:

* Add/define flag in pkg/flags - done
* Update the [envconf.Config](https://github.com/kubernetes-sigs/e2e-framework/blob/33c3d5efc189c44ff7dbc283a361481a264c8005/pkg/envconf/config.go#L33) with the context config (i.e. `kubeContext` field).
* Next update [envconf.NewFromFlags](https://github.com/kubernetes-sigs/e2e-framework/blob/33c3d5efc189c44ff7dbc283a361481a264c8005/pkg/envconf/config.go#L62:6) so that the environment can be created with the new flag
* Next, add a pair of accessor methods `Config.WithKubernetesContext` and `Config.KubernetesContext` similar to how it is done for Namespace [here](https://github.com/kubernetes-sigs/e2e-framework/blob/33c3d5efc189c44ff7dbc283a361481a264c8005/pkg/envconf/config.go#L144)
* Update [`Config.NewClient()`](https://github.com/kubernetes-sigs/e2e-framework/blob/11aa681198560751f4694b0ae613d580bbb50012/pkg/envconf/config.go#L111) to create the klient.Client with the cluster name, when the `kubeConfig` is provided.

It sounds complicated, but we are here to assist you in getting this in.

cc @harshanarayana

maruina:
> So, my next question to you, do you have to pass your cluster context name as a CLI argument ? You could use an environment variable (i.e. KUBE_CONTEXT) and read it from your test, then apply that method. That way you won't even have to wait for PR or a code change.

Correct me if I'm wrong, but in the test we usually call `cfg.NewClient()` where cfg is a `*envconf.Config`. So I would still need to make some changes, or I'm not 100% sure what you mean with `apply that method`

Regarding the flag, thank you very much for your input, I will follow up with your suggestions

maruina:
@vladimirvivien I think that I followed your suggestions up to this point
>Update [Config.NewClient()](https://github.com/kubernetes-sigs/e2e-framework/blob/11aa681198560751f4694b0ae613d580bbb50012/pkg/envconf/config.go#L111) to create the klient.Client with the cluster name, when the kubeConfig is provided.

When I look at the code, we call [client, err := klient.NewWithKubeConfigFile(c.kubeconfig)](https://github.com/kubernetes-sigs/e2e-framework/blob/11aa681198560751f4694b0ae613d580bbb50012/pkg/envconf/config.go#L116) to create a client with a specific kubeconfig, so my understanding is that I need to provide a new method or change `NewWithKubeConfigFile`. 

Do you want me to check if `c.kubecontext` is not an empty string inside [NewClient()](https://github.com/kubernetes-sigs/e2e-framework/blob/11aa681198560751f4694b0ae613d580bbb50012/pkg/envconf/config.go#L111)? The check we do for the kubeconfig is later in the code, when we call [conf.New(filePath)](https://github.com/kubernetes-sigs/e2e-framework/blob/11aa681198560751f4694b0ae613d580bbb50012/klient/conf/config.go#L35)

So I can call [NewWithContextName(fileName, context string)](https://github.com/kubernetes-sigs/e2e-framework/blob/11aa681198560751f4694b0ae613d580bbb50012/klient/conf/config.go#L47) but I probably need something like 
```
client, err := klient.NewWithKubeConfigFileAndKubeContext(c.kubeconfig, c.kubecontext)
```
but again, it feels a little too convoluted. What do you think?

vladimirvivien:
> So I can call [NewWithContextName(fileName, context string)](https://github.com/kubernetes-sigs/e2e-framework/blob/11aa681198560751f4694b0ae613d580bbb50012/klient/conf/config.go#L47) but I probably need something like
> 
> ```
> client, err := klient.NewWithKubeConfigFileAndKubeContext(c.kubeconfig, c.kubecontext)
> ```

@maruina I think you are getting closer. In `Config.NewClient` you could check to see if the context is non-empty then create the client as shown below. You can  create a new constructor function `klient.NewWithContextName(file, string)` that uses `klient/conf.NewWithContextName()` to build the `klient.Client`

```go
func (c *Config) NewClient() (klient.Client, error) {
	if c.client != nil {
		return c.client, nil
	}

        if c.kubecontext != "" {
            client, err := klient.NewWithContextName(c.kubeconfig, c.kubeContext)
	    if err != nil {
		return nil, fmt.Errorf("envconfig: client failed: %w", err)
	    } 
           c.client = client
        }else{
	   client, err := klient.NewWithKubeConfigFile(c.kubeconfig)
	   if err != nil {
  	       return nil, fmt.Errorf("envconfig: client failed: %w", err)
	   }
	   c.client = client
       }

	return c.client, nil
}
```

harshanarayana:
@vladimirvivien could we not just offload all calls from `NewClient` to the `klient.NewWithContextNam` and handle the cases there where the context is set of not? Since that is really a `klient` related case than the `Config` is it not ?

harshanarayana:
@maruina

Sorry in advance for the really lengthy comment. 


If you check the args generated for the binaries of e2e-framework, it already has a `-kube-context` arg. May be we should consider reusing that instead of adding a new one?

```bash
❯ ./flags.test --help 2>&1 | grep -A2 kube-cont
  -kube-context string
    	Used to provide a custom KubeContext override for the file passed to the --kubeconfig
  -kubeconfig string
```

https://github.com/kubernetes-sigs/e2e-framework/blob/cd133e286ac7059c7743561cb6441542d60674f4/klient/conf/config.go#L112-L122

I think we should reuse this and modify the workflow as bit in terms of how the client is created. 

```golang
func New(fileName string) (*rest.Config, error) {
	// if filename is not provided assume in-cluster-config
	if fileName == "" {
		return rest.InClusterConfig()
	}

	if kubeContext := ResolveClusterContext(); kubeContext != "" {
		return NewWithContextName(fileName, kubeContext)
	}
	
	// create the config object from k8s config path
	return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
		&clientcmd.ClientConfigLoadingRules{ExplicitPath: fileName}, &clientcmd.ConfigOverrides{}).ClientConfig()
}
```

This should keep the code change to really minimal ? cc @vladimirvivien  




If we are to add a new one, I think here are a few suggestion. 

```golang
func NewWithKubeConfigFileAndContext(kubeConfig string, kubeContext string) (Client, error) {
	if kubeContext == "" {
		return NewWithKubeConfigFile(kubeConfig)
	}
	cfg, err := conf.NewWithContextName(kubeConfig, kubeContext)
	if err != nil {
		return nil, err
	}
	return New(cfg)
}
```
I think we can add a function like so on the `pkg/klient/client.go` and call that from `pkg/envconf/config.go` like following

```golang
func (c *Config) NewClient() (klient.Client, error) {
	if c.client != nil {
		return c.client, nil
	}

	client, err := klient.NewWithKubeConfigFileAndContext(c.kubeconfig, c.kubeContext)
	if err != nil {
		return nil, fmt.Errorf("envconfig: client failed: %w", err)
	}
	c.client = client

	return c.client, nil
}
```
This will keep the isolation intact to a good extend. 

Also a few other things to consider, 

1. If the `--kubeconfig` arg is not provided but a `--kube-context` is provided, then we should probably throw an error ?
2. I think it is better to name the arg `--kube-context` instead of `--kubecontext` for readability. 
3. Using proper camel cases in the code. Please consider using `kubeContext` instead of `kubecontext`
4. Since we are looking into it, I think the signature of `func NewWithContextName(fileName, context string) (*rest.Config, error) ` might be better changed a bit on the variable name ? The second one reads `context` which is really conflicting with the default package. Might be better to rename this `kubeContext` ?

maruina:
I still pushed the change so if we are happy with that, we can test it :)
I can always revert to the previous state

harshanarayana:
> @harshanarayana from which branch did you try the flags? I'm trying with the latest main but it seems I can't see that `kube-context` flag


Oh no! I think I made a mistake. I'm terribly sorry about that.. I was on your PR branch and had modified the arg name.. I totally forgot about that when I ran the help command.. That's how the variable showed up there... You still have to add that arg yourself..

maruina:
@harshanarayana no worries. I pushed again the flag, and I was able to target a specific cluster using 
```
go test . -v -args --kubeconfig ~/.kube/config --labels category=mylabel --kube-context mycontext
```

pkg/flags/flags.go
```diff
@@ -38,6 +38,7 @@ const (
 	flagDryRunName              = "dry-run"
 	flagFailFast                = "fail-fast"
 	flagDisableGracefulTeardown = "disable-graceful-teardown"
+	flagKubeContext             = "kube-context"
```
harshanarayana, line 41: I think it would be nice to have a document added in the examples indicating how to use this flag.

pkg/flags/flags.go
```diff
@@ -247,6 +259,10 @@ func ParseArgs(args []string) (*EnvFlags, error) {
 		flag.BoolVar(&disableGracefulTeardown, disableGracefulTeardownFlag.Name, false, disableGracefulTeardownFlag.Usage)
 	}
 
+	if flag.Lookup(kubeContextFlag.Name) == nil {
```
harshanarayana, line 262: I think this should be checked against the presence of the `-kubeconfig` argument as well.  Without that, this argument doesn't really hold much of a values. Is it not ?

harshanarayana (changes requested):

harshanarayana:
@maruina  Leaving a few minor comments. Once done, let us see if @vladimirvivien  is okay with this approach. 

/cc @ShwethaKumbla

pkg/flags/flags.go
maruina: I thought we were using the default location at `~/.kube/config`, but I'll add a check

vladimirvivien (commented):
Apologies @maruina and @harshanarayana for the delay in response. Thank you @harshanarayana for the thoughtful feedback in this PR. I think it is ready. @maruina please squash the commits and I will take one last look to get this in.

maruina:
@vladimirvivien squash done, please ping me if there is anything else I can do

pkg/flags/flags.go
```diff
@@ -264,6 +280,10 @@ func ParseArgs(args []string) (*EnvFlags, error) {
 		panic(fmt.Errorf("--fail-fast and --parallel are mutually exclusive options"))
 	}
 
+	if kubeconfig == "" && kubeContext != "" {
+		panic(fmt.Errorf("--kube-context requires the --kubeconfig option"))
```
vladimirvivien, line 284: While this is true, it is not necessary to have the --kubeconfig flag for this to work. This is because if [`ResolveKubeConfigFile`](https://github.com/kubernetes-sigs/e2e-framework/blob/db4e9fe7e23d2602c34d1fe3853625e754b1fa41/klient/conf/config.go#L67) is called during the construction of the client instance, it will attempt to resolve the kubeconfig automatically.  

Maybe that check should be moved [above](https://github.com/kubernetes-sigs/e2e-framework/pull/197/files#diff-3923bb7d5d40d33c81ef1c40584ce3ef3f81c7c6bb1ffb2adfc842b745533a7cR41) when setting the flags up.

klient/conf/config.go
```diff
@@ -38,6 +38,11 @@ func New(fileName string) (*rest.Config, error) {
 		return rest.InClusterConfig()
 	}
 
+	// set the desired kube context
+	if kubeContext := ResolveClusterContext(); kubeContext != "" {
```
vladimirvivien, line 42: This check should also check to see if kubeConfig is available (either via the --kubeconfig flag value or via [`ResolveKubeConfigFile`](https://github.com/kubernetes-sigs/e2e-framework/blob/db4e9fe7e23d2602c34d1fe3853625e754b1fa41/klient/conf/config.go#L67). If the value of KubeConfig is not provided, this should error out. See comment below for detail.

vladimirvivien (commented):
@maruina I have left one comment (split in two places) for one last check. Then I think it's ready for a merge.

klient/conf/config.go
maruina: > If the value of KubeConfig is not provided, this should error out.

But if we are not passing a kubeconfig we should hit https://github.com/kubernetes-sigs/e2e-framework/blob/db4e9fe7e23d2602c34d1fe3853625e754b1fa41/klient/conf/config.go#L37 because we're assuming in-cluster-config

My understanding is that we call `ResolveKubeConfigFile` in https://github.com/kubernetes-sigs/e2e-framework/blob/db4e9fe7e23d2602c34d1fe3853625e754b1fa41/examples/real_cluster/main_test.go#L38 and then we construct the config from that file, without using any flag.

I think that you're saying that we could use `ResolveKubeConfigFile` to automatically get the kubeconfig *and* we could use the flags to also do additional configurations, am I correct? But `NewFromFlags()` doesn't take any argument, so I think I'm missing something

vladimirvivien: Hi @maruina . You are correct in your last thinking. 
We could force --kubeconfig to be provided when --context is used or (as i tried to suggest)
* Check for value of kubeConfig
* if empty, call ResolveKubeConfigFile
* if still empty, then error out

Also, if there is a better place to make that check work, you can put it where it works best.

examples/flags/README.md
```diff
@@ -28,6 +28,7 @@ There are several supported flags (for more accurate list, see package `pkg/flag
 * `features`
 * `labels`
 * `kubeconfig`
+* `kube-context`
```
vladimirvivien, line 31: @maruina one more minor suggestion
Let's call the flag just `context`. This is to stay aligned with the naming convention (usually) followed by other tools in the kubernetes ecosystem.

vladimirvivien (commented):
This is taking shape nicely. Thanks for all the work.

examples/flags/README.md
maruina: Yep, done. That's why I was suggesting it in the fist place

klient/conf/config.go
maruina: Ok, I think I've done that. Please have a look :)

klient/conf/config.go
```diff
@@ -33,14 +34,33 @@ var DefaultClusterContext = ""
...
+		resolvedKubeConfigFile = fileName
+	}
+
+	// if resolvedKubeConfigFile is still empty, assume in-cluster config
+	if resolvedKubeConfigFile == "" {
+		if context == "" {
+			return rest.InClusterConfig()
+		}
+		// if in-cluster can't use the --context flag
+		return nil, errors.New("cannot use --context flag without a kubeconfig file")
```
vladimirvivien, line 53: I would reword the error to just mention "cluster context" instead of "--context flag" since context can come in programmatically at this point in the code, not necessarily via flag.

vladimirvivien (commented):
@maruina I only have one nit comment. After this I think it's ready.

klient/conf/config.go
```diff
@@ -33,14 +34,33 @@ var DefaultClusterContext = ""
 // New returns Kubernetes configuration value of type *rest.Config.
 // filename is kubeconfig file
 func New(fileName string) (*rest.Config, error) {
-	// if filename is not provided assume in-cluster-config
+	var resolvedKubeConfigFile string
+	context := ResolveClusterContext()
```
harshanarayana, line 38: Please do not shadow the context here with the default package name. Use `kubeContext` instead

klient/conf/config.go
harshanarayana: The argument is `--kube-context` right? Can you please update the name of the arg as well

klient/conf/config.go
maruina: Ops you are right, sorry. Fixed 🤦

klient/conf/config.go
maruina: @harshanarayana , no as @vladimirvivien said in the previous comment, we switched back to `--context` to be consistent with other Kubernetes CLI. For example `kubectl` uses `--context`.

harshanarayana: Sorry, I totally missed that part of the discussion. @maruina Thanks

pkg/flags/flags.go
```diff
@@ -277,6 +293,7 @@ func ParseArgs(args []string) (*EnvFlags, error) {
 		dryRun:                  dryRun,
 		failFast:                failFast,
 		disableGracefulTeardown: disableGracefulTeardown,
+		context:                 context,
```
harshanarayana, line 296: @vladimirvivien I know you dissed about keep this consistent with the kubectl in terms of args. Which Is perfectly fine. But I think we should use the internal variables/attributes in such a way that it won't conflict with the `context` package. That can lead to a lot of confusion. Do you think it will be ok to at least name the internal attributes/variables as `kubeContext` but retain the flag argument as `--context` ?

harshanarayana (commented):
Sorry, for doing this in bits. Leaving a couple of more nitpicks. Missed them last time as I was on the phone.

pkg/flags/flags.go
maruina: Not @vladimirvivien but I think you have a point and I can change that. I didn't think about the `context` package and the shadowing at all when writing this.

vladimirvivien: @maruina @harshanarayana Yes, I thought about `context` causing confusion with context package. So, what I am suggesting is just the name of the flag being called `context`. However internally, the variable name should be kubeContext (everywhere it's used).  Hope that makes sense.

vladimirvivien: I think after that, this will be ready to go and we can have a release soon.

pkg/envconf/config.go
```diff
@@ -44,6 +44,7 @@ type Config struct {
 	dryRun                  bool
 	failFast                bool
 	disableGracefulTeardown bool
+	kubeContext             string
```
maruina, line 47: @vladimirvivien and @harshanarayana, even if we're parsing directly from flags in `ResolveClusterContext()`, I thought about adding the flag to the `Config` object for consistency.

If you think it's not necessary, I'm happy to drop this change

maruina:
I tested the squashed commit locally and I can run some tests to a target context with
```
❯ go test . -v -args --labels category=my-category -context my-context
```
The nice thing is that I don't have to specify the kubeconfig flag anymore.

If we pass an invalid file, for example
```
❯ go test . -v -args --labels category=my-category -context my-context --kubeconfig /tmp/foo
F0214 22:56:20.312484   71123 env.go:369] Setup failure: create namespace func: envconfig: client failed: cannot use a cluster context without a valid kubeconfig file
FAIL	github.com/my-repo/e2eframework	0.362s
FAIL
```
That is because we bubble up `kubeconfig` as an empty string and we don't get the associated error - in this case the file doesn't exist.

pkg/flags/flags.go
harshanarayana: @vladimirvivien  @maruina  Works for me.

vladimirvivien (commented):
I apologize, I meant to send this earlier this week. After this, I think it's ready for a merge.

maruina:
@vladimirvivien updated the code :)

vladimirvivien (approved):
Thanks @maruina for this awesome contribution.

harshanarayana:
@maruina 

```bash
go: downloading github.com/josharian/intern v1.0.0
E0217 13:49:12.378771    9620 env.go:352] Recovering from panic and running finish actionsenvconfig: client failed: open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory
program not built with -cover
ok  	sigs.k8s.io/e2e-framework/examples/crds	41.446s
=== RUN   TestListPods
=== RUN   TestListPods/pod_list
=== RUN   TestListPods/pod_list/pods_from_kube-system
    k8s_test.go:33: envconfig: client failed: open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory
--- FAIL: TestListPods (0.00s)
    --- FAIL: TestListPods/pod_list (0.00s)
        --- FAIL: TestListPods/pod_list/pods_from_kube-system (0.00s)
FAIL
coverage: [no statements]
FAIL	sigs.k8s.io/e2e-framework/examples/custom_env_funcs	38.357s
E0217 13:50:14.534158   15810 env.go:352] Recovering from panic and running finish actionsenvconfig: client failed: open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory
program not built with -cover
ok  	sigs.k8s.io/e2e-framework/examples/decoder	20.[72](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_e2e-framework/197/pull-e2e-framework-test/1626578895165722624#1:build-log.txt%3A72)6s
```

Seems like new tests are breaking the tests

klient/conf/config.go
```diff
@@ -33,14 +34,34 @@ var DefaultClusterContext = ""
...
+	kubeContext := ResolveClusterContext()
+
+	// resolve the kubeconfig file
 	if fileName == "" {
-		return rest.InClusterConfig()
+		resolvedKubeConfigFile = ResolveKubeConfigFile()
+		if fileName != "" {
+			resolvedKubeConfigFile = fileName
+		}
+	}
```
harshanarayana, line 46: Sorry, I missed this one 🤦 

This can't be happening right ? Your first check is `fileName == ""` and inside the lop you have a `fileName != ""` without first setting the `fileName` to a new values.

maruina:
@vladimirvivien I think that I actually broke something, let me see where the problem is because I think I tested everything before adding the two new methods so I'm not sure how I broke things

harshanarayana (changes requested):

klient/conf/config.go
harshanarayana: @maruina  I think I know why. Check the above one. 

```golang
// resolve the kubeconfig file
	resolvedKubeConfigFile = fileName
	if fileName == "" {
		resolvedKubeConfigFile = ResolveKubeConfigFile()
	}
```

Should be like above.

maruina: Yes, of course 🤦 
Tests are passing locally now.

vladimirvivien:
@maruina ping me when you think you got it fixed.

maruina:
@vladimirvivien all the tests are passing locally, can you please retest?

harshanarayana (approved):
;lgtm

vladimirvivien:
Thanks @maruina !!

maruina:
Thanks both of you for your patience and help

Issue #202, opened 2023-02-08
Use custom flags


maruina:
Hello there.

I have a use case where I need to pass some extra flags to the the CLI for configuring some of my tests.

What do you think about adding and `extraArgs` field
```
type EnvFlags struct {
    [...]
    extraArgs map[string]string
}
```
or something similar? Then we can access those args from the tests by calling something like
```
myValue := cfg.ExtraArgs("myKey")
```
Or we can create something similar to LabelsMap and call it `type ExtraArgsMap map[string]string` and add some methods to it
What do you think?

harshanarayana:
Is is not already possible using the `flag` package ?

```bash
❯ ./flags.test --help 2>&1 | grep -A2 my-custom
  -my-custom-flag string
    	My Custom Flag Integration
```

```diff
diff --git a/examples/flags/flags_test.go b/examples/flags/flags_test.go
index 064e722..4f46fce 100644
--- a/examples/flags/flags_test.go
+++ b/examples/flags/flags_test.go
@@ -18,6 +18,7 @@ package flags

 import (
        "context"
+       "flag"
        "os"
        "testing"

@@ -54,7 +55,11 @@ var test env.Environment
 // Then, execute the test:
 //
 //  ./flags.test --assess en
+var myCustomFlag string
+
 func TestMain(m *testing.M) {
+       flag.StringVar(&myCustomFlag, "my-custom-flag", "", "My Custom Flag Integration")
+
        // create config from flags (always in TestMain because it calls flag.Parse())
        cfg, err := envconf.NewFromFlags()
        if err != nil {
```

You can store this outside on a package of your own and access them when you see fit. Is there a reason you want these flags to be integrated with the `EnvFlags` itself ?

harshanarayana:
```diff
diff --git a/examples/flags/flags_test.go b/examples/flags/flags_test.go
index 064e722..238cf64 100644
--- a/examples/flags/flags_test.go
+++ b/examples/flags/flags_test.go
@@ -18,6 +18,8 @@ package flags

 import (
        "context"
+       "flag"
+       "fmt"
        "os"
        "testing"

@@ -54,13 +56,19 @@ var test env.Environment
 // Then, execute the test:
 //
 //  ./flags.test --assess en
+var myCustomFlag string
+
 func TestMain(m *testing.M) {
+       // Add custome flag
+       flag.StringVar(&myCustomFlag, "my-custom-flag", "", "My Custom Flag Integration")
+
[...]

PR #203, opened 2023-02-13
Add documentation about passing custom flags

maruina:
Fix https://github.com/kubernetes-sigs/e2e-framework/issues/202


examples/flags/README.md
```diff
@@ -68,3 +68,37 @@ To get additional verbose logs
 ```shell
 ./flags.test --assess es --v 2
 ```
+
+You can also pass additional custom flags
+
+```shell
+./flags.test --myflag myvalue
+```
+
```
harshanarayana, line 77: Would you mind turning this code into an actual go file that can be run or update the existing flag test itself to include this behavior or customizing the flags? That would be really cool to have a working example that you can run as well instead of just read.

vladimirvivien: Same suggestion. Or it may even makes more sense to create a new example under `examples/custom-flags/`, with README.md and a go example to show how to do it.

maruina: Done, does this look better?

examples/custom_flags/custom_flags_test.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+var (
+	test         env.Environment
+	myCustomFlag string
+)
+
+func TestMain(m *testing.M) {
+	flag.StringVar(&myCustomFlag, "my-custom-flag", "", "my custom flag for my tests")
+	// create config from flags (always in TestMain because it calls
```
harshanarayana, line 39: `TestMain` or `init` handler of the package. Both will do, But should be done before the `envconf.NewFromFlags()` is called. That is all. Mind updating the comment here ?

maruina: Done. Do you also want me to update https://github.com/kubernetes-sigs/e2e-framework/blob/db4e9fe7e23d2602c34d1fe3853625e754b1fa41/examples/flags/flags_test.go#L58 since I copied from that?

PR #210, opened 2023-03-01
feat: add env.TestFunc type alias

### LINKED ISSUE #199: Expose `TestEnvFunc` similarly as it's done with `FeatureEnvFunc` and `EnvFunc`
Currently `FeatureEnvFunc` and `EnvFunc` are exposed in https://github.com/kubernetes-sigs/e2e-framework/blob/main/pkg/env/env.go#L37-L41 to alias the underlying internal types.

This unfortunately is not being done for `TestEnvFunc`.

One can work around this simply by defining it on their own:

```
type TestEnvFunc = func(context.Context, *envconf.Config, *testing.T) (context.Context, error)
```

but it doesn't feel right.

This would be beneficial to e.g. allow end users mapping step funcs to test funcs

```
type TestEnvFunc = func(context.Context, *envconf.Config, *testing.T) (context.Context, error)

func IntoTestEnvFunc(stepFn features.Func) TestEnvFunc {
	return func(ctx context.Context, e *envconf.Config, t *testing.T) (context.Context, error) {
		return stepFn(ctx, t, e), nil
	}
}
```

### Proposed solution

Expose `TestEnvFunc`.

pmalek:
Closes #199

### DIFF
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -38,6 +38,7 @@ type (
 	Environment = types.Environment
 	Func        = types.EnvFunc
 	FeatureFunc = types.FeatureEnvFunc
+	TestFunc    = types.TestEnvFunc
 )
 
 type testEnv struct {


vladimirvivien (approved):
Thanks for the contribution

PR #224, opened 2023-03-20
Select flag and labels logic change

### LINKED ISSUE #218: Make --labels check if any of the labels exists, not all
Currently the `--labels` and `--skip-labels` flag will create a `map[string][]string` and based on the key and value pairs ,decide whether a feature is to be run or not. 

The problem is that if I run using `--labels "key1=value1,key2=value2"` with the below features

```go

f1 := features.New(....).WithLabel("key1","value1").Assess("something one", ... ).Feature()

f1 := features.New(....).WithLabel("key2","value2").Assess("something two", ... ).Feature()

testenv.Test(t,f1,f2)
```

Neither of the tests run.

- `f1` will not run as it does not have `key2=value2`
- `f2` will nto run as it does not have `key1=value1`

What I would expect is  for the framework to check if any of the passed labels exist , in this case it would run both.

This behaviour is weird given that for `--skip-labels` , it skips based on the existence of any single label.

Not sure if we should modify the behaviour of `--labels` or have another flag for this situation ( the latter is probably the better idea )

I can have a PR ready soon if people agree.

pranavb23:
Resolves #218 .

The naming could be changed a bit but the idea is the same. I've tested the changes on a sample test locally.

Will update the test files soon.

This PR adds a new flag `--select-labels` which will AND the labels ( similar to the previous behaviour of `--labels` ) and change the `--labels` flag behaviour to OR the labels.

### DIFF
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -524,7 +524,8 @@ func (e *testEnv) requireProcessing(kind, testName string, requiredRegexp, skipR
 			}
 		}
 
-		skip = true
+		// if there are no labels passed through --labels then we can directly set skip to false and continue, the loop wont run anyway
+		skip = len(e.cfg.Labels()) != 0
 	checkLabels:
 		for key, vals := range e.cfg.Labels() {
 			for _, v := range vals {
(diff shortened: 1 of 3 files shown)


pranavb23:
@harshanarayana the tests need to be updated, I'm a bit busy atm. Will find time to update the tests asap and get this ready.

PR #225, opened 2023-03-20
Updates for dependabot and ci job

cpanato:
- Updates for dependabot and ci job

/assign @vladimirvivien

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,20 +14,21 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@24cb9080177205b6e8c946b17badbe402adc938f # v3.4.0
         with:
           fetch-depth: 0
 
       - name: Set up Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@4d34df0c2316fe8122ab82dc22947d607c0c91f9 # v4.0.0
         with:
-          go-version: 1.17.x
+          go-version: '1.20'
+          check-latest: true
 
       - name: Run GoReleaser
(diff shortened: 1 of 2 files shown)


ShwethaKumbla (approved):
LGTM

.github/workflows/release.yml
```diff
@@ -14,20 +14,21 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@24cb9080177205b6e8c946b17badbe402adc938f # v3.4.0
```
vladimirvivien, line 17: Hi @cpanato ! Curious why the hash instead of the the tag?

vladimirvivien (commented):
Left a question.

.github/workflows/release.yml
cpanato: it is a best practice to use the hash instead of the tag, the tag is mutable and can be changed, but the hash not :)
and we pin to a specific release. also dependabot can update that and also update the version comment :)


hope that clarifies

vladimirvivien: Yes. Thanks!

vladimirvivien:
@cpanato I saw that too. I plan to check it out as well. Thanks.

harshanarayana:
@vladimirvivien  @cpanato  I noticed the same too. it seems to be the same set of test that are failing since March 25th. I am checking locally to see what might be causing it

cpanato:
> @vladimirvivien @cpanato I noticed the same too. it seems to be the same set of test that are failing since March 25th. I am checking locally to see what might be causing it

fix here: https://github.com/kubernetes-sigs/e2e-framework/pull/229

Issue #228, opened 2023-04-01
Noticing a large number of flaky Tests `TestHelmChartRepoWorkflow`


harshanarayana:
```bash
=== RUN   TestHelmChartRepoWorkflow/Repo_based_helm_chart_workflow/Deployment_is_running_successfully
    helm_test.go:68: failed waiting for the Deployment to reach a ready state
=== RUN   TestHelmChartRepoWorkflow/Repo_based_helm_chart_workflow/run_Chart_Tests
--- FAIL: TestHelmChartRepoWorkflow (302.30s)
    --- FAIL: TestHelmChartRepoWorkflow/Repo_based_helm_chart_workflow (302.30s)
        --- FAIL: TestHelmChartRepoWorkflow/Repo_based_helm_chart_workflow/Deployment_is_running_successfully (300.01s)
        --- PASS: TestHelmChartRepoWorkflow/Repo_based_helm_chart_workflow/run_Chart_Tests (0.15s)
```
<img width="1002" alt="Screenshot 2023-04-01 at 4 00 20 PM" src="https://user-images.githubusercontent.com/1404867/229280961-9c420ba4-8490-4538-a64f-c74c8f02cadc.png">

A large amount of tests runs have been failing with this one since March 25th.

cpanato:
fix: https://github.com/kubernetes-sigs/e2e-framework/pull/229

PR #229, opened 2023-04-01
update helm test

cpanato:
Fixes: https://github.com/kubernetes-sigs/e2e-framework/issues/228

/assign @vladimirvivien @ShwethaKumbla @harshanarayana

### DIFF
--- a/examples/third_party_integration/helm/helm_test.go
+++ b/examples/third_party_integration/helm/helm_test.go
@@ -56,7 +56,7 @@ func TestHelmChartRepoWorkflow(t *testing.T) {
 		Assess("Deployment is running successfully", func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
 			deployment := &appsv1.Deployment{
 				ObjectMeta: v1.ObjectMeta{
-					Name:      "nginx-nginx-ingress",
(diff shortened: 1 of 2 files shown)


examples/third_party_integration/helm/helm_test.go
```diff
@@ -56,7 +56,7 @@ func TestHelmChartRepoWorkflow(t *testing.T) {
 		Assess("Deployment is running successfully", func(ctx context.Context, t *testing.T, config *envconf.Config) context.Context {
 			deployment := &appsv1.Deployment{
 				ObjectMeta: v1.ObjectMeta{
-					Name:      "nginx-nginx-ingress",
+					Name:      "nginx-nginx-ingress-controller",
```
harshanarayana, line 59: @vladimirvivien  should we have a dedicated chart pushed to the helm registry that is just meant for our test and use that instead in order to avoid failures like this one for example in the future ?

cc @cpanato

cpanato: sounds good, i can do that if you all want

vladimirvivien:
@cpanato & @harshanarayana  thanks for an awesome job!!

PR #231, opened 2023-04-05
pin helm chart to a specific version

cpanato:
Follow up of https://github.com/kubernetes-sigs/e2e-framework/pull/229/files#r1155138390

but instead, publish our own lets pin to a specific version that is working
will be too much work and does worth publishing a test one for testing IMHO

/assign @vladimirvivien @harshanarayana


vladimirvivien (approved):
LGTM
This is a simple, but good idea to pin to a specific ver.

PR #239, opened 2023-04-16
Add support for kwok cluster simulator

reetasingh:
Fixes https://github.com/kubernetes-sigs/e2e-framework/issues/214

- Adding support for creation and deletion of `kwok` cluster which includes installing `kwokctl`
- Adding env function for creation and deletion of kwok cluster
- Adding example

PR #246, opened 2023-04-18
cluster provider interface definition

### LINKED ISSUE #245: Enhancement: Define Cluster Provider Interfaces for ease of extension
Until recently, the e2e framework had just one way to create a local cluster and that was using `kind`. 

However, with the changes that @reetasingh  has been working on for https://github.com/kubernetes-sigs/e2e-framework/pull/239, there is now `kwok` support. This bring up a point that there is not a standard interface that defines the cluster provider (We should also find a suitable name for this. Cluster provider might not be the right one) so that we can easily extend the `envfunc` without having to duplicate them for each of the clusters. 

This started in the discussion of the above PR under https://github.com/kubernetes-sigs/e2e-framework/pull/239#discussion_r1168259000

```golang
type clusterOpts func(c ClusterProvider)

type ClusterProvider interface {
	WithVersion(version string) ClusterProvider
	WithOpts(opts ...clusterOpts) ClusterProvider
	Create(args ...string) (string, error)
	GetKubeConfig() string
	GetKubectlContext() string
	ExportLogs(dest string) error
	Destroy() error
	LoadDockerImages(image string) error
	LoadImageArchive(archivePath string) error
}
```

Above is a possible example of what the interface might look like (This is in no way the what we want to use eventually but just for the same of discussion). 

Defining above will help us do a few things. 

1. Parameterise the `envfuncs` helpers for `kind/kwok` so that they are just one set of functions and and invoked via the interface types 
2. It also lets extend this easily in the future to add additional provides such as `k3s` or `kubeadm` for example

harshanarayana:
Closes #245 

# Why? 

Lately we have been seeing a lot of traction towards adding more cluster providers for performing e2e tests other than `kind` such as `kwok`  `k3d` etc.

Every provider defining their own set of functions and handlers will soon complicate the integration points and unnecessarily duplicate the need for certain generic `env.Func`.. This PR aims to provide a standard interface to be implemented by the cluster providers so that they can be seamlessly integrated into the framework


support/types.go
```diff
@@ -0,0 +1,65 @@
...
+	// Destroy is used to cleanup a cluster brought up as part of the test workflow
+	Destroy() error
+
+	// LoadDockerImage is used to load a set of Docker images to the cluster via the cluster provider native workflow
+	LoadDockerImage(image string) error
+
+	// LoadImageArchive is used to provide a mechanism where a tar.gz archive containing the docker images used
+	// by the services running on the cluster can be imported and loaded into the cluster prior to the execution of
+	// test if required.
+	LoadImageArchive(archivePath string) error
```
jbpratt, line 64: I'm still trying to get familiar with `kwok` but I'm not sure if there is a concept of this so these may be specific enough to `kind` to leave off? Imagining a (albeit complex) `kubeadm` or OpenShift (I guess there is `oc import-image`) provider, I'm not sure if these would apply either. Curious to hear your thoughts!

reetasingh: +1, did not find an option in `kwokctl` to do this.

support/provider.go
```diff
@@ -0,0 +1,37 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package support
+
+type ClusterProvider string
+
+const (
+	Kind ClusterProvider = "Kind"
```
vladimirvivien, line 22: I am not sure this package constant belong in here (maybe in support/kind/kind.go).

support/provider.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+var ProviderRegistry map[ClusterProvider]E2EClusterProvider
+
+func GetClusterProvider(provider ClusterProvider) E2EClusterProvider {
+	return ProviderRegistry[provider]
+}
+
+func RegisterProvider(name ClusterProvider, provider E2EClusterProvider) {
+	ProviderRegistry[name] = provider
```
vladimirvivien, line 32: This should be protected in with a sync.Mutex.

vladimirvivien (commented):
Great start @harshanarayana

pkg/envfuncs/kind_funcs.go
```diff
@@ -28,31 +28,32 @@ import (
...
 
-// CreateKindCluster returns an env.Func that is used to
+// CreateCluster returns an env.Func that is used to
 // create a kind cluster that is then injected in the context
 // using the name as a key.
 //
 // NOTE: the returned function will update its env config with the
 // kubeconfig file for the config client.
-func CreateKindCluster(clusterName string) env.Func {
+func CreateCluster(clusterName string) env.Func {
```
ShwethaKumbla, line 54: Is this renaming really required?  
As we are going to add other cluster providers (ex: kwok)  under this package it should also have all these functions with the provider name and I think it will ease user's understanding of which provider has been used to create the cluster.

ShwethaKumbla (changes requested):

support/types.go
harshanarayana: @vladimirvivien  @jbpratt @reetasingh  

I totally agree that not all providers need to have this capabilities or will provide these capabilities. I kept that under this just to simplify things and avoid having to type assert things and then invoke handlers selectively in some places if we decide to use these functions in the future. The providers that do not need to support this can just have a dummy handler defined that is a no-op. Or for cases of `kubeadm` for example, it can `docker/crictl pull` the image to the cluster as part of the load if someone decides to take that route. 

Is there a real advantages of taking this out of this type ?

support/provider.go
harshanarayana: Ack

pkg/envfuncs/kind_funcs.go
harshanarayana: Yes and no. The reason behind refactoring this is to make this a bit more generic. That part is not yet sorted out. 

The Idea is `CreateCluster` will take a `ClusterProvider` type as argument and then fetch the provider using that name and then trigger the workflow from there. 

There will be bunch of alias handler created to `CreateKindCluster` that proxies the call to `CreateCluster` to retain compatibility with existing users. 

We can annotate those function as deprecated and slowly ask consumer to move towards the generic handlers across the board. Also the file will be renamed as `provider_funcs.go` may be so that there are not many direct references to kind on the code base

harshanarayana:
@jbpratt @reetasingh @ShwethaKumbla @vladimirvivien  Sorry I have been holding hostage a bunch of PRs because of this work still being pending. I have been out sick for a while and just getting back to the routine. Pushing a bunch of changes to this as we speak

support/types.go
```diff
@@ -0,0 +1,65 @@
...
+	// WithOpts provides a way to customize the options that can be used while setting up the
+	// cluster using the providers such as kind or kwok or anything else. These helpers can be
+	// leveraged to setup arguments or configuration values that can be provided while performing
+	// the cluster bring up
+	WithOpts(opts ...ClusterOpts) E2EClusterProvider
+
+	// Create Provides an interface to start the cluster creation workflow using the selected provider
+	Create(args ...string) (string, error)
+
+	CreateWithConfig(kindConfigFile string) (string, error)
```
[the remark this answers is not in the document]
harshanarayana: Ack

vladimirvivien:
Hi all 👋 
I think this is going to be an import PR for future integration with cluster provides (i.e Kind, K3D, etc)
@harshanarayana  How is this PR doing? What else is needed to get it over the line.

(sorry I have been missing in action on this)

harshanarayana:
@vladimirvivien nothing more pending on my end for this for now. Waiting for a review

jbpratt:
hey @harshanarayana @vladimirvivien, I'm curious what you think about taking a generics approach to this problem https://go.dev/play/p/bEYNyo9lucH

harshanarayana:
> hey @harshanarayana @vladimirvivien, I'm curious what you think about taking a generics approach to this problem https://go.dev/play/p/bEYNyo9lucH

That's exactly what this PR is implementing isn't it?

jbpratt:
> > hey @harshanarayana @vladimirvivien, I'm curious what you think about taking a generics approach to this problem https://go.dev/play/p/bEYNyo9lucH
> 
> That's exactly what this PR is implementing isn't it?

As far as I can tell, you aren't using generics but maybe I'm missing it? You maintain a map of providers and take the provider name as a parameter to fetch it out of the map.

https://github.com/kubernetes-sigs/e2e-framework/blob/575c4a4a57736e9d33a5bd2907c410751dbf6d95/pkg/envfuncs/provider_funcs.go#L85-L87

https://github.com/kubernetes-sigs/e2e-framework/blob/575c4a4a57736e9d33a5bd2907c410751dbf6d95/support/provider.go#L25

My recommendation is slightly different in that we depend on type safety rather than a map at runtime

harshanarayana:
> > > hey @harshanarayana @vladimirvivien, I'm curious what you think about taking a generics approach to this problem https://go.dev/play/p/bEYNyo9lucH
> > 
> > That's exactly what this PR is implementing isn't it?
> 
> As far as I can tell, you aren't using generics but maybe I'm missing it? You maintain a map of providers and take the provider name as a parameter to fetch it out of the map.
> 
> https://github.com/kubernetes-sigs/e2e-framework/blob/575c4a4a57736e9d33a5bd2907c410751dbf6d95/pkg/envfuncs/provider_funcs.go#L85-L87
> 
> https://github.com/kubernetes-sigs/e2e-framework/blob/575c4a4a57736e9d33a5bd2907c410751dbf6d95/support/provider.go#L25
> 
> My recommendation is slightly different in that we depend on type safety rather than a map at runtime

Ah I see.. Sorry I totally missed that you were referring to generics.. We can do that with generics. But that limits the go version for someone adding this as a dependency on their libraries though.. Wouldn't it?

harshanarayana:
@jbpratt in second thought, I think that would be a great idea.. Do we need to worry about supporting go version less that 1.18? If not, I'll switch the PR to generics

jbpratt:
> @jbpratt on second thought, I think that would be a great idea.. Do we need to worry about supporting go version less that 1.18? If not, I'll switch the PR to generics

It does limit dependents to be >=1.18 but our go.mod already does that https://github.com/kubernetes-sigs/e2e-framework/blob/main/go.mod#L3 so I don't think it should be a problem! Everything older than 1.19 is EOL anyway.

Sounds good! Thank you for your work on this

harshanarayana:
```go
func waitForControlPlane(client klient.Client) error {
	r, err := resources.New(client.RESTConfig())
	if err != nil {
		return err
	}
	selector, err := metav1.LabelSelectorAsSelector(
		&metav1.LabelSelector{
			MatchExpressions: []metav1.LabelSelectorRequirement{
				{Key: "component", Operator: metav1.LabelSelectorOpIn, Values: []string{"etcd", "kube-apiserver", "kube-controller-manager", "kube-scheduler"}},
			},
		},
	)
	if err != nil {
		return err
	}
	// an e2e provider cluster with one control-plane node will have 4 pods running the core apiserver components
	err = wait.For(conditions.New(r).ResourceListN(&v1.PodList{}, 4, resources.WithLabelSelector(selector.String())))
	if err != nil {
		return err
	}
	selector, err = metav1.LabelSelectorAsSelector(
		&metav1.LabelSelector{
			MatchExpressions: []metav1.LabelSelectorRequirement{
				{Key: "k8s-app", Operator: metav1.LabelSelectorOpIn, Values: []string{"kindnet", "kube-dns", "kube-proxy"}},
			},
		},
	)
	if err != nil {
		return err
	}
	// an e2e provider cluster with one control-plane node will have 4 k8s-app pods running networking components
	err = wait.For(conditions.New(r).ResourceListN(&v1.PodList{}, 4, resources.WithLabelSelector(selector.String())))
	if err != nil {
		return err
	}
	return nil
}
```

I am trying to find a generic way to setup this helper function.  Since this is not really generic the way it works today, i am wonder if I can add an extra interface function that provides a list of selectors and we can use that to perform this wait function ?

harshanarayana:
@jbpratt  With generics, do you want me to turn the existing `kind` cluster type to value receiver or use the pointer receiver ? Which one would work better ?

jbpratt:
> @jbpratt With generics, do you want me to turn the existing `kind` cluster type to value receiver or use the pointer receiver ? Which one would work better ?

Hey @harshanarayana, sorry about that I realized last night this would probably be a problem. I had a friend show me this which allows us to use the pointer receivers https://go.dev/play/p/qdFCw5w42DB I'd just be worried our structs are too large

harshanarayana:
> > @jbpratt With generics, do you want me to turn the existing `kind` cluster type to value receiver or use the pointer receiver ? Which one would work better ?
> 
> Hey @harshanarayana, sorry about that I realized last night this would probably be a problem. I had a friend show me this which allows us to use the pointer receivers https://go.dev/play/p/qdFCw5w42DB I'd just be worried our structs are too large otherwise

@jbpratt I did this already. About to push the changes

pkg/envfuncs/provider_funcs.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+// Deprecated: This handler has been deprecated in favor of CreateClusterWithConfig which can now accept
+// support.ClusterProvider type as input in order to setup the cluster using right providers
+func CreateKindClusterWithConfig(clusterName, image, configFilePath string) env.Func {
+	return CreateClusterWithConfig[kind.Cluster](clusterName, configFilePath, kind.WithImage(image))
+}
+
+// Deprecated: This handler has been deprecated in favor of DestroyCluster
+func DestroyKingCluster(name string) env.Func {
```
jbpratt, line 91:
```suggestion
func DestroyKindCluster(name string) env.Func {
```

harshanarayana: 🤣  that was silly typo. let me fix it

harshanarayana:
@jbpratt @reetasingh @ShwethaKumbla @vladimirvivien @cpanato PTAL.. I think I've done all the changes I had In my mind.

pkg/envfuncs/provider_funcs.go
```diff
@@ -0,0 +1,272 @@
...
+
+type clusterNameContextKey string
+
+// Following placeholder interface types are defined purely to make the pointer receiver work seemlessly
+// for the generic handlers. Without this, making the pointer receiver is really painful or we have to
+// turn providers into value receivers and that doesn't really work well for the current implementation
+// we have for kind provider. xref: https://t.ly/GDxOB
+
+// Every time you add a new provider that implements the support.E2EClusterProvider interface, that
+// needs to be listed here in this interface type with properly or'ing that with the existing
```
reetasingh, line 43: nit: typo ` properly or'ing`

reetasingh (approved):
LGTM @harshanarayana 
supporting new provider will be easy with this. I can see it working well for kwok

pkg/envfuncs/provider_funcs.go
```diff
@@ -0,0 +1,272 @@
...
+
+// Deprecated: This handler has been deprecated in favor of DestroyCluster
+func DestroyKindCluster(name string) env.Func {
+	return DestroyCluster(name)
+}
+
+// Deprecated: This handler has been deprecated in favor of ExportClusterLogs
+func ExportKindClusterLogs(name, dest string) env.Func {
+	return ExportClusterLogs(name, dest)
+}
```
vladimirvivien, line 98: I would keep these in the kind_funcs.go file

pkg/envfuncs/provider_funcs.go
```diff
@@ -0,0 +1,272 @@
...
+		kubecfg, err := k.Create()
+		if err != nil {
+			return ctx, err
+		}
+
+		// update envconfig  with kubeconfig
+		cfg.WithKubeconfigFile(kubecfg)
+
+		// stall, wait for pods initializations
+		if err := waitForControlPlane(p, cfg.Client()); err != nil {
```
vladimirvivien, line 119: Suggestion: waiting for cluster control plane probably should be a function provided by the cluster provider.

pkg/envfuncs/provider_funcs.go
```diff
@@ -0,0 +1,272 @@
...
+			return ctx, err
+		}
+
+		// store entire cluster value in ctx for future access using the cluster name
+		return context.WithValue(ctx, clusterNameContextKey(clusterName), k), nil
+	}
+}
+
+func waitForControlPlane(provider support.E2EClusterProvider, client klient.Client) error {
```
vladimirvivien, line 156: I applaud this attempt of making this function convenient for the provider writers. However, I think it makes a lot of assumptions about how a provider works and how it waits for the control plane.  Again, I would suggest making it the responsibility of the provider to know how to wait for its own control plane.

pkg/envfuncs/provider_funcs.go
```diff
@@ -0,0 +1,272 @@
...
+// we have for kind provider. xref: https://t.ly/GDxOB
+
+// Every time you add a new provider that implements the support.E2EClusterProvider interface, that
+// needs to be listed here in this interface type with properly or'ing that with the existing
+// values.
+//
+//	type provider interfae {
+//			kind.Cluster | k3d.Cluster | kwok.Cluster
+//	}
+type provider interface {
```
vladimirvivien, line 49: I am a bit confused why this interface extends a kind.Cluster type.
Is `type provider` kind specific?

pkg/envfuncs/provider_funcs.go
```diff
@@ -0,0 +1,272 @@
...
+	kind.Cluster
+}
+
+type providerRef[T any] interface {
```
vladimirvivien, line 53: What is the role of the providerRef type ?

pkg/envfuncs/provider_funcs.go
```diff
@@ -0,0 +1,272 @@
...
+
+// CreateCluster returns an env.Func that is used to
+// create an E2E provider cluster that is then injected in the context
+// using the name as a key.
+//
+// NOTE: the returned function will update its env config with the
+// kubeconfig file for the config client.
+func CreateCluster[T provider, P providerRef[T]](clusterName string) env.Func {
+	return func(ctx context.Context, cfg *envconf.Config) (context.Context, error) {
+		var p P = new(T)
```
vladimirvivien, line 108: If `T` is assignable to `P`, why not just use a single type to represent provider?
Maybe I am missing something.

vladimirvivien (commented):
@harshanarayana Thank you for a great contribution, it is shaping up nicely. However, I left some questions regarding the design, specifically the use of generic types provider and providerRef.

Let's continue to work on it because I think it will be a great addition.

pkg/envfuncs/provider_funcs.go
harshanarayana: I have kept this here for compatibility purpose alone as someone might already have this import in their workflow. Moving this would break that. Can we retain this here for now and eventually get rid of it down the line instead of moving this to kind_funcs.go entirely ?

pkg/envfuncs/provider_funcs.go
harshanarayana: I was thinking just that. That would simplify things for sure. I will move this out.

pkg/envfuncs/provider_funcs.go
harshanarayana: https://github.com/kubernetes-sigs/e2e-framework/pull/246#discussion_r1232437324

I am going to move this into the provider.

pkg/envfuncs/provider_funcs.go
harshanarayana: These hacked up two interface mechanism is the only way to make pointer recievers work with go generics right now. I have not found a much simpler solution :( 

xref: https://t.ly/GDxOB

harshanarayana: >  Is type provider kind specific?

Not it is not. That is why I also made sure to add comments. Say we add a new Provider like kwok,  then this line has to be updated to `kind.Cluster | kwok.Cluster`

harshanarayana:
@vladimirvivien  Thanks for the review. I have updated the changes suggested. I am open to suggestion on if there is a way to simplify the generics part of pointer receiver. But I have not found one so, I will leave that as is for now. PTAL when you can.

klient/decoder/decoder_test.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package decoder
+package decoder_test
```
harshanarayana, line 17: Had to modify some of these to avoid dependency import cycles. These are standard go convention anyway. Switched only the selective bits to this format.

harshanarayana:
Screenshot 2023-06-17 at 1 13 33 AM

@vladimirvivien  @cpanato  any idea why this might be happening ? Tests have gotten really flaky again in the last couple of days. Even the dependabot once have been failing. Most of them seem to be failing because of ` Setup failure: failed to create kind cluster: exit status 1 : exit status 1`.

pkg/envfuncs/provider_funcs.go
vladimirvivien: Ok, thanks for explaining. So that's the only way to implement pointer receiver with genericized types.

klient/decoder/decoder_test.go
vladimirvivien: While reviewing yesterday, I wanted to bring up that issue.  (Even if not in this PR) We should take a look at reorg the package layout.   

For instance, pkg/envfunc should not depend support/kind (but I understand that's how we did at first). 
All pkg packages should be depended on by other external packages (not the other way around).  Otherwise we will keep getting into these tricky circular dependencies.

harshanarayana: @vladimirvivien  +1. I totally agree. I believe we need to reorg the packages a bit to cleanup some of these. I remember we had this discussion as part of a different Git Issue but did not take that forward then. Let us open a Git Tracker issue and sort that out right away. I also want to start using the `_test` convention in the packages for the Tests if that is all right with everyone, Makes it more cleaner in terms of organizing and avoid a lot of dependency issues too.

pkg/envfuncs/provider_funcs.go
harshanarayana: @vladimirvivien  yes for now. We will revisit this again if there is a better way to achieve this for sure. I will ask around with some go experts that I know to see if they can help me to check if there is a better way.

pkg/envfuncs/provider_funcs.go
vladimirvivien: @harshanarayana I meant move it to its own file envfuncs/kind_funcs.go as before, in the same packge.

pkg/envfuncs/provider_funcs.go
vladimirvivien: > These hacked up two interface mechanism is the only way to make pointer recievers work with go generics right now. I have not found a much simpler solution :(
> 
> xref: https://t.ly/GDxOB

@harshanarayana As the link points out, this Go approach is a bit clunky and is a workaround.
Until you pointed it out, I hadn't paid much attention to such issues with Go generics.

pkg/envfuncs/provider_funcs.go
```diff
@@ -0,0 +1,242 @@
...
+	return ExportClusterLogs(name, dest)
+}
+
+// CreateCluster returns an env.Func that is used to
+// create an E2E provider cluster that is then injected in the context
+// using the name as a key.
+//
+// NOTE: the returned function will update its env config with the
+// kubeconfig file for the config client.
+func CreateCluster[T provider, P providerRef[T]](clusterName string) env.Func {
```
vladimirvivien, line 100: What if `CreateCluster` takes an E2EClusterProvider as a type constraint  as follows?

```go
func CreateCluster[P E2EClusterProvider](provider P, clusterName string) env.Func { 
...
}
```

I think it will work simpler, right ? Since your type constraint will be an interface you should be able to use type parameter that only needs to implement the type constraint as follows:

```go
func CreateKindCluster(clusterName string) env.Func {
	return CreateCluster[kind.Cluster](p kind.Cluster, clusterName)
}
```

I could be wrong. But it's worth looking into if it helps simplifying the implementation.

vladimirvivien (commented):
Hey, take a look a the comments.

pkg/envfuncs/provider_funcs.go
harshanarayana: @vladimirvivien  this is what me and @jbpratt  were discussing earlier 

xref: https://github.com/kubernetes-sigs/e2e-framework/pull/246#issuecomment-1586079640
xref: https://github.com/kubernetes-sigs/e2e-framework/pull/246#issuecomment-1585486410

If we are to use the interface directly, like this, it will work as value receiver, which will break our workflow where we set the kubeconfig files to the context of the cluster provider as part of the cluster create. We have to make that as pointer receiver and the hack I have put up right now is the only way to do that as far as I know.

pkg/envfuncs/provider_funcs.go
harshanarayana: Done

harshanarayana:
@vladimirvivien  Thanks for the quick review. I have updated the changes based on your comments for kind_funcs.go. I have also ported the early implementation of the `kwok` from the original PR for reference. Please let me know if you want me to move that out and have a different PR later on.

pkg/envfuncs/provider_funcs.go
harshanarayana: @vladimirvivien  I was speaking to @merovius on a Gopher channel and Axel suggested something that is much simper that what I have been doing. https://go.dev/play/p/C7RbbQYQzyN

This seem to do the job. May be this will simplify things for us.

harshanarayana: And the other suggestion that made a lot of sense was to just avoid generics entirely and keep it simple. Which I think would be the best thing to do here since generics is not really making this any better. cc @jbpratt

harshanarayana: All right.. Let me refactor and cleanup the generics and use a simple approach like we had before

support/kwok/kwok.go
```diff
@@ -0,0 +1,224 @@
...
+	k.rc = cfg
+	cs, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return nil, err
+	}
+	k.client = cs
+	return k.client, nil
+}
+
+func (k *Cluster) LoadDockerImage(_ string) error {
```
vladimirvivien, line 139: I am a bit concerned about returning nil on a noop method.
We should make sure this is documented as a noop and/or returned an error.

jbpratt:
> @vladimirvivien Thanks for the quick review. I have updated the changes based on your comments for kind_funcs.go. I have also ported the early implementation of the `kwok` from the original PR for reference. Please let me know if you want me to move that out and have a different PR later on. PTAL when you can.

@harshanarayana Do we plan to remove the kwok implementation from this PR following https://github.com/kubernetes-sigs/e2e-framework/pull/239#issuecomment-1601917798 ? Not much reason in fixing it up if so.

harshanarayana:
@jbpratt  Yes. I will move that out of this PR and push a commit. Was doing just that as we speak

support/kwok/kwok.go
harshanarayana: @vladimirvivien  I have removed this from this PR for now. So depending on which one gets merged first out of this and #239, we can handle the kwok bit accordingly.

klient/decoder/main_test.go
```diff
@@ -14,35 +14,39 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package decoder
+package decoder_test
```
vladimirvivien, line 17: @harshanarayana I am coming back here so it doesn't fall the radar. `decoder_test`  means our dependency graph has an issue. Let's open a PR (if you haven't done so) for package re-org to avoid this situation.

vladimirvivien (commented):
@harshanarayana PR looks great. Thanks for rewriting the API to use the interface. It is simpler (I think) to reason about.

klient/decoder/main_test.go
harshanarayana: @vladimirvivien We will move that work of refactoring into a different PR. That might break some existing imports and others for the existing users. So better to track that as a different item?

vladimirvivien: Agreed.

support/kwok/kwok.go
```diff
@@ -122,89 +138,125 @@ func (k *Cluster) Create(args ...string) (string, error) {
...
+	if err := k.findOrInstallKwokCtl(); err != nil {
+		return err
 	}
-	// installing kwok means installing kwokctl binary
-
-	if e.Prog().Avail("kwokctl") == "" {
-		log.V(4).Info(`kwokctl not found, installing version @%s`, kwokVersion)
-		if err := k.installKwokCtl(e); err != nil {
-			return err
+	for _, component := range []string{"audit", "etcd", "kube-apiserver", "kube-controller-manager", "kube-scheduler", "kwok-controller", "prometheus"} {
```
harshanarayana, line 172: @reetasingh  PTAL, this is the closest to export log that I could find with the `kwokctl`.

harshanarayana: Actually, there might be a better way to do this with https://github.com/kubernetes-sigs/kwok/pull/630/files

Let me see if we can find a way to use that

vladimirvivien: @harshanarayana does the code pulls/install kwokctl bin similar to kind ?

harshanarayana: @vladimirvivien  Yes it does. You can also pass a custom bin file to it from non standard path just like https://github.com/kubernetes-sigs/e2e-framework/pull/287

support/types.go
```diff
@@ -0,0 +1,100 @@
...
+	// failures better after the fact.
+	ExportLogs(dest string) error
+
+	// Destroy is used to cleanup a cluster brought up as part of the test workflow
+	Destroy() error
+
+	// LoadDockerImage is used to load a set of Docker images to the cluster via the cluster provider native workflow
+	// Not every provider will have a mechanism like this/need to do this. So, providers that do not have this support
+	// can just provide a no-op implementation to be compliant with the interface
+	LoadDockerImage(image string) error
```
wzshiming, line 73:
```suggestion
	LoadImage(image string) error
```

Images may not just be for Docker.

support/types.go
```diff
@@ -0,0 +1,100 @@
...
+	// CreateWithConfig is used to provide a mechanism where cluster providers that take an input config
+	// file and then setup the cluster accordingly. This can be used to provide input such as kind config
+	CreateWithConfig(configFile string) (string, error)
+
+	// GetKubeconfig provides a way to extract the kubeconfig file associated with the cluster in question
+	// using the cluster provider native way
+	GetKubeconfig() string
+
+	// GetKubeCtlContext is used to extract the kubectl context to be used while performing the operation
+	GetKubeCtlContext() string
```
wzshiming, line 60: IMO, `GetKubectlContext` or `GetKubeconfigContext` may be more appropriate

support/types.go
```diff
@@ -0,0 +1,100 @@
...
+	// and is expected to return an error if the control plane doesn't stabilize. If the provider being implemented
+	// does not have a clear mechanism to identify the Control plane readiness or is not required to wait for the control
+	// plane to be ready, such providers can simply add a no-op workflow for this function call.
+	// Returning an error message from this handler will stop the workflow of e2e-framework as returning an error from this
+	// is considered as  failure to provision a cluster
+	WaitForControlPlane(klient.Client) error
+
+	// KubernetesClient is a helper function used to provide an instance of kubernetes.Interface which can then be used
+	// to interact with Provider clusters in a generic way if the existing resources handlers/helpers are not enough
+	KubernetesClient() (kubernetes.Interface, error)
```
wzshiming, line 99: Maybe we also need a `dynamic.Interface`, `*rest.Config` and `*rest.RESTClient`

support/types.go
```diff
@@ -0,0 +1,100 @@
...
+	WithPath(path string) E2EClusterProvider
+
+	// WithOpts provides a way to customize the options that can be used while setting up the
+	// cluster using the providers such as kind or kwok or anything else. These helpers can be
+	// leveraged to setup arguments or configuration values that can be provided while performing
+	// the cluster bring up
+	WithOpts(opts ...ClusterOpts) E2EClusterProvider
+
+	// Create Provides an interface to start the cluster creation workflow using the selected provider
+	Create(args ...string) (string, error)
```
wzshiming, line 49:
```suggestion
	Create(ctx context.Context, args ...string) (string, error)
```

support/types.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// CreateWithConfig is used to provide a mechanism where cluster providers that take an input config
+	// file and then setup the cluster accordingly. This can be used to provide input such as kind config
+	CreateWithConfig(configFile string) (string, error)
```
wzshiming, line 53:
```suggestion
	CreateWithConfig(ctx context.Context, configFile string) (string, error)
```

support/types.go
```diff
@@ -0,0 +1,100 @@
...
+	// GetKubeCtlContext is used to extract the kubectl context to be used while performing the operation
+	GetKubeCtlContext() string
+
+	// ExportLogs is used to export the cluster logs via the cluster provider native workflow. This
+	// can be used to export logs from the cluster after test failures for example to analyze the test
+	// failures better after the fact.
+	ExportLogs(dest string) error
+
+	// Destroy is used to cleanup a cluster brought up as part of the test workflow
+	Destroy() error
```
wzshiming, line 68:
```suggestion
	Destroy(ctx context.Context) error
```

support/types.go
```diff
@@ -0,0 +1,100 @@
...
+	SetDefaults() E2EClusterProvider
+
+	// WaitForControlPlane is a helper function that can be used to indiate the Provider based cluster create workflow
+	// that the control plane is fully up and running. This method is invoked after the Create/CreateWithConfig handlers
+	// and is expected to return an error if the control plane doesn't stabilize. If the provider being implemented
+	// does not have a clear mechanism to identify the Control plane readiness or is not required to wait for the control
+	// plane to be ready, such providers can simply add a no-op workflow for this function call.
+	// Returning an error message from this handler will stop the workflow of e2e-framework as returning an error from this
+	// is considered as  failure to provision a cluster
+	WaitForControlPlane(klient.Client) error
```
wzshiming, line 95:
```suggestion
	WaitForControlPlane(ctx context.Context, klient.Client) error
```

support/types.go
harshanarayana: Hmm. Wondering if it would be a better option to just provide `rest.Config` from here and let the consumers build the client for their liking? That will also keep the external facing interfaces limited.

support/types.go
harshanarayana: I am wondering if we can just add the ctx argument to all the functions defined by the interface that are doing some work other than just returning an attribute value ?

support/kwok/kwok.go
reetasingh: @harshanarayana I think we can say the supported version is v0.3.0 and use `kwokctl export logs`

support/types.go
```diff
@@ -0,0 +1,120 @@
...
+	// that the control plane is fully up and running. This method is invoked after the Create/CreateWithConfig handlers
+	// and is expected to return an error if the control plane doesn't stabilize. If the provider being implemented
+	// does not have a clear mechanism to identify the Control plane readiness or is not required to wait for the control
+	// plane to be ready, such providers can simply add a no-op workflow for this function call.
+	// Returning an error message from this handler will stop the workflow of e2e-framework as returning an error from this
+	// is considered as  failure to provision a cluster
+	WaitForControlPlane(ctx context.Context, client klient.Client) error
+}
+
+type E2EClusterProviderWithK8sHelper interface {
```
vladimirvivien, line 102: Hi @harshanarayana 
I am not sure this separate type is needed, just to return Kubneretes clients.
Would having Type `E2EClusterProvider` return an env.Config (which has the client) be enough, as it's done in other part of the code?  If not, would can we return something from which the client can be built rest.Config or something?

support/utils/command.go
```diff
@@ -0,0 +1,84 @@
...
+
+var commandRunner = gexe.New()
+
+// FindOrInstallGoBasedProvider check if the provider specified by the pPath executable exists or not.
+// If it exists, it returns the path with no error and if not, it uses the `go install` capabilities to
+// install the provider and setup the required binaries to perform the tests. In case if the install
+// is done by this helper, it will return the value for installed binary as provider which can then
+// be set in the in the invoker to make sure the right path is used for the binaries while invoking
+// rest of the workfow after this helper is triggered.
+func FindOrInstallGoBasedProvider(pPath, provider, module, version string) (string, error) {
```
vladimirvivien, line 35: Seeing this method sort of remind me of the kubetest2 code. It seems that we are approaching recreating kubetest2 and its provider model. 

It maybe worth taking a look there to see if there is anything we can leverage (if any).  We do have specific needs, but if we can avoid recreating the wheels completely, it's worth checking it out.

support/utils/kubernetes.go
```diff
@@ -0,0 +1,40 @@
...
+package utils
+
+import (
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/e2e-framework/klient/conf"
+)
+
+func FetchKubernetesClients(config string) (cfg *rest.Config, kubernetesClient kubernetes.Interface, dynamicClient dynamic.Interface, err error) {
```
vladimirvivien, line 26: I think we should think about simplifying this, may be break it into two separate functions to return the K8s client needed (i.e. `FetchDynamicClient` and `FetchTypedClient` or something similar).

vladimirvivien (commented):
@harshanarayana Given the size of this one, we knew that it was going to be a while before it gets in. I see progress, but I left some comments as well.

support/types.go
harshanarayana: I am fine to return just the `rest.Config`. Which was my original plan too. https://github.com/kubernetes-sigs/e2e-framework/pull/246#discussion_r1252678904

I thought having a way to easily access this might help some use cases, but then again, the advantages are not too many. I will just add a function that can return `rest.Config` so that it can be used accordingly for any needs.

support/utils/command.go
harshanarayana: Yes. The idea is largely inspired by their model of providers. I don't think I found anything on kubetest2 that we can reuse here directly. Do you have anything specific in mind with regards to re-usability ?

support/utils/kubernetes.go
harshanarayana: I will get rid of this entirely in favor of keeping the interface simple and have the interface return the `rest.Config` that one can use to create whatever client they need.

support/utils/command.go
harshanarayana: https://github.com/kubernetes-sigs/kubetest2/blob/e61130e0c75e2dc357d46230993909955a95a646/pkg/types/types.go#L91 

What I found interesting was that, it has defined interfaces much differently. This lets the providers have optional capabilities. But I don't think we need to be that granular in our case here since we want most of the common functionality to be provided by all providers. Using that convention, I have moved the image load helpers out to their own interface so that providers that do not provide that feature can simply ignore that.

vladimirvivien (approved):
This looks great. We can refine if needed at some point later.
Thanks @harshanarayana 🎉

PR #262, opened 2023-05-19
feat: Display helm command errors

bradbeam:
This change captures stderr from helm commands. 
This changes the current behavior of returning `exit status 1` to now include the reason for the failure, ex:

```
Error: INSTALLATION FAILED: rendered manifests contain a resource that already exists...
```


vladimirvivien (approved):
Thank you @bradbeam

PR #278, opened 2023-06-19
handle conversion of resource handler options properly between metav1 and controller-runtime

### LINKED ISSUE #254: Delete options removed by controller runtime intermediary client
Expected: Delete options passed into `Resources.Delete` are used by controller-runtime
Actual: Delete options passed into `Resources.Delete` are removed by controller-runtime

Calling like
```
err = r.Delete(ctx, &job, resources.WithDeletePropagation(string(metav1.DeletePropagationBackground)))
```

The cr delete options are set so that all fields are empty except for `Raw`.
See https://github.com/kubernetes-sigs/e2e-framework/blob/ce6e190f552cc0c082d92c41b2e9dff43c31d9ad/klient/k8s/resources/resources.go#L139

Controller runtime constructs its own DeleteOptions from these options here https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/typed_client.go#L80

This ends up calling https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/options.go#L300 which will result in still only `Raw` being set

then `AsDeleteOptions` is called at the end of `typedClient.Delete` which sets the fields of `Raw` directly from the member fields of the `DeleteOptions` https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/options.go#L273, those fields are not set so it ends up zeroing out the fields of `Raw`.

It seems like controller runtime does not expect `Raw` to be set and overrides all fields set in it.

Confirmed the following workaround works:
```
deletionPropagation := metav1.DeletePropagationBackground
err = r.GetControllerRuntimeClient().Delete(ctx, &job, &cr.DeleteOptions{PropagationPolicy: &deletionPropagation})
```

harshanarayana:
Fixes #254

**NOTE:**

I have made the changes to all the options used by this resources handler since they can all run into similar issues as reported in #254

### DIFF
--- a/klient/k8s/resources/resources.go
+++ b/klient/k8s/resources/resources.go
@@ -24,6 +24,8 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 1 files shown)


harshanarayana:
@vladimirvivien @ryanmcnamara  @negz PTAL when possible.

vladimirvivien (approved):
@harshanarayana thanks for getting this done quickly.
LGTM

PR #281, opened 2023-06-21
enable linters of examples and update golangci-lint version

### LINKED ISSUE #280: `make verify` workflow doesn't validate the `examples/**` components.
The defaults used by the `golangci-lint` for `--skip-dirs-use-default` includes `- (^|/)examples($|/)` and `- (^|/)third_party($|/)`

As a side effect of this, none of the code that we have in `examples/**` is getting linted. We should probably fix this. Here is what needs to be done to enable this. 

1. Modify the existing `.golangci.yml` to include the following
```yaml
run:
  concurrency: 6
  deadline: 5m
  skip-dirs-use-default: false
  skip-dirs:
    - (^|/)vendor($|/)
    - (^|/)testdata($|/)
    - (^|/)Godeps($|/)
    - (^|/)builtin($|/)
```
2. Once we are done with that, a series of linter checks will fail on the examples
3. Disable the following linters on the examples dir 
```yaml
    - path: examples/*
      linters:
        - dupl
        - staticcheck
        - gocritic
        - stylecheck
        - misspell
```

4. Enable `revive` to detect diplicate imports across the components
```yaml
  revive:
    confidence: 0.01
    rules:
      - name: duplicated-imports
        severity: error
        disabled: false
```

Opening this as a catch all fix based on what was see in https://github.com/kubernetes-sigs/e2e-framework/pull/279

harshanarayana:
Enabling linters on the `example/**` components so that they can be linted alongside the other packages codes.

Along with that, this PR also bumps up the `golangci-lint` version and sets-up the `depguard` linter config for allowed packages under tests and main

Closes #280

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -30,7 +30,6 @@ linters:
   enable:
     - asciicheck
     - bodyclose
-    - depguard
     - dogsled
     - dupl
     - errcheck
@@ -62,26 +61,6 @@ linters:
     - whitespace
 
 linters-settings:
-  # Addig the following section based on https://github.com/golangci/golangci-lint/issues/3877
-  depguard:
-    rules:
-      main:
-        files:
-          - $all
-          - "!$test"
-        allow:
-          - $gostd
-          - github.com/vladimirvivien
-          - k8s.io
-          - sigs.k8s.io
-      test:
-        files:
-          - "$test"
-        allow:
-          - $gostd
-          - github.com/vladimirvivien
-          - k8s.io
-          - sigs.k8s.io
   revive:
     confidence: 0.01
     rules:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,6 +2,12 @@
 run:
   concurrency: 6
   deadline: 5m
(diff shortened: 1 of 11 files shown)


.golangci.yml
```diff
@@ -47,6 +62,32 @@ linters:
     - whitespace
 
 linters-settings:
+  # Addig the following section based on https://github.com/golangci/golangci-lint/issues/3877
+  depguard:
```
cpanato, line 66: i would drop this, this is a bit expensive to maintain and not sure what values this adds

for example we dropped this check in the sigstore org

harshanarayana: @cpanato  I added this because not having this was breaking my local run of `make verify`  :( Should I disable that plugin entirely instead ?

cpanato: yes, remove that check

harshanarayana:
@cpanato  Changes done to cleanup the `depguard` check. PTAL when possible.

cpanato (approved):
thanks
/lgtm

PR #255, opened 2023-05-10
Add Flux support to install resources from git source:

matrus2:
Hello,

This a proposal and MVP for integration e2e framework with Flux. This one include:
* create and delete GitRepository,
* create and delete Kustmozation,
* example of how to use the functionality. 

Please comment & review. Once first part is done we can focus to provide all other functionalities from Flux e.g. Helm repository reconciliation support.

PR #283, opened 2023-06-23
Add templates for flux example

matrus2:
Hi, 

This is hello-world go application template, which will be referenced by flux Kustomization example in #255. This needs to merged first otherwise tests won't pass in corresponding PR. 

cc: @harshanarayana @vladimirvivien

### DIFF
--- a/examples/third_party_integration/flux/template/deployment.yaml
+++ b/examples/third_party_integration/flux/template/deployment.yaml
@@ -1,0 +1,20 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: hello-app
+  namespace: hello-world
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: hello-app
+  template:
+    metadata:
+      labels:
+        app: hello-app
+    spec:
+      containers:
+        - name: hello-app
+          image: matrus2hub/go-hello-world@sha256:2c4bd6771dbb88db8d8294245d2937f797ff30708d0647aec564e8cd7851c2b5
+          ports:
+            - containerPort: 8888
--- a/examples/third_party_integration/flux/template/deployment.yaml
+++ b/examples/third_party_integration/flux/template/deployment.yaml
@@ -15,6 +15,6 @@ spec:
     spec:
       containers:
         - name: hello-app
-          image: matrus2hub/go-hello-world@sha256:2c4bd6771dbb88db8d8294245d2937f797ff30708d0647aec564e8cd7851c2b5
+          image: nginx:1.25.1
           ports:
-            - containerPort: 8888
+            - containerPort: 80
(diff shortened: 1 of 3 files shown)


examples/third_party_integration/flux/template/deployment.yaml
```diff
@@ -0,0 +1,20 @@
...
+    matchLabels:
+      app: hello-app
+  template:
+    metadata:
+      labels:
+        app: hello-app
+    spec:
+      containers:
+        - name: hello-app
+          image: matrus2hub/go-hello-world@sha256:2c4bd6771dbb88db8d8294245d2937f797ff30708d0647aec564e8cd7851c2b5
```
harshanarayana, line 18: @cpanato  @vladimirvivien  do we have an image coming from the kubernetes-sigs that we can use here instead of this ? It would be great if there is one. I did a quick search through the `test-infra` and did not find one. May be we can just use a simple upstream nginx image instead ? Sorry for being very picky on the source of the images.

matrus2: @harshanarayana You are completely right, I switched to nginx instead.

vladimirvivien: Along the same line of questioning, @cpanato @harshanarayana, since nginx is on Dockerhub, will repo-testing run into pull throttling issues, causing it to fail ?

harshanarayana: We might very well run into that. But I did a bit of digging around in the test-infra and `kubernetes-sigs` or and I found a combination of these images used already. 

Most of the nginx images used in the `k-sigs` are from direct dockerhub. But I did find a few other options 

1. registry.k8s.io/nginx-slim:0.8 (https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/blob/1712c870e2ed773c3de0be8138502d8bb82c39d0/examples/kubernetes/multishare/maxshares/demo-ss-512.yaml#L21)
2. mcr.microsoft.com/oss/nginx/nginx:1.19.5 (https://github.com/kubernetes-sigs/blob-csi-driver/blob/9c74b4a3d14683a26958af9cd2a4f65656f5569e/deploy/example/statefulset.yaml#L20)

I believe it should be ok to use the default dockerhub images so long as we are using the same docker registry credentials in prow for this repo jobs as the one being used for others. cc @cpanato  @vladimirvivien

harshanarayana: @vladimirvivien  Let us get this one merged. We can revisit the images from a different registry again if we start seeing flaky tests because of the rate limiting.

vladimirvivien: Thanks @harshanarayana for looking into this.

/lgtm
/approve

PR #286, opened 2023-06-25
enable using kind/helm binaries from non standard location

### LINKED ISSUE #282: Support explicitly specifying which CLI binaries to invoke
The e2e-framework offers packages to help spin up Kind clusters and install Helm charts. These packages wrap the `kind` and `helm` CLI tools. (There may be other wrapped CLI tools but I haven't encountered them yet.)

The framework seems to look for these tools with hardcoded binary names (`kind`, `helm`) in `$PATH`. If `kind` isn't available the framework attempts to install it using `go install`. If `helm` isn't available the framework returns an error.

In @crossplane our `Makefile` handles installation of tools like Kind and Helm, but it doesn't install them in `$PATH`. We can probably hack around this when running the E2E tests (e.g. add a dir to `$PATH` and symlink to the tools from that dir) but it would be ideal to be able to pass the path to the binaries as flags.

harshanarayana:
Closes #282

### DIFF
--- a/support/kind/kind.go
+++ b/support/kind/kind.go
@@ -31,6 +31,7 @@ import (
 var kindVersion = "v0.17.0"
 
 type Cluster struct {
+	path        string
 	name        string
 	e           *gexe.Echo
 	kubecfgFile string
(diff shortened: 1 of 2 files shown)


cpanato (approved):
Cool

/lgtm

PR #265, opened 2023-05-25
Fix race condition when executing parallel tests

maruina:
This should fix https://github.com/kubernetes-sigs/e2e-framework/issues/258 and https://github.com/kubernetes-sigs/e2e-framework/issues/216

The issue seems to be we're embedding a context inside `testenv` and when we run multiple tests in parallel we write back to that context from those multiple go routines, causing a data race.

The change I'm proposing is to be explicit and pass a context explicitly to `Run`, `Test` and `TestInParallel`. If the majority of the users care about context propagation because they use it to pass data between tests or their different phases, I think we should provide a thread safe data structure to do so.

I can't reproduce the issue anymore with my example code and the `pkg/env` tests are passing.

PR #291, opened 2023-06-29
Fix race condition when executing parallel tests, keeping context

phisco:
This is another approach to solve the same problem as #265 and was developed with the help of @maruina.

This patch makes the context in the testEnv thread safe, but implements the following logic:

- If `Test` is used, the same logic of now applies. So the context is passed to all features being run sequentially and is then reinjected into the environment.
- If `TestInParallel` is used (and parallel tests are enabled), the context is first populated by the `BeforeEachTest` functions and then a dedicated child of the parent context is provided to each test running in parallel, which is then discarded and only the original parent context is passed to the `AfterEachTest` functions and back into the environment. This is actually fixing a race condition in the previous implementation of `TestInParallel`.
- If `t.Parallel` is used to run one or more features either using `Test` or `TestInParallel`, the same logic above applies, with the exception that multiple tests will be accessing the same context in the BeforeEachTest and AfterEachTest functions, and given that the update of the context is done by substitution and is not atomic, the order, and therefore the outcome, is non-deterministic. So it's up to the user to define a dedicated context for each test in this case.

### DIFF
--- a/hack/test-go.sh
+++ b/hack/test-go.sh
@@ -38,5 +38,5 @@ cd "${REPO_ROOT}"
 
 # Ensure -p=1 to avoid packages running concurrently which may all try and install kind at the same time or race for
 # use of the kind binary.
-GO111MODULE=on go test -v -p=1 -timeout="${TEST_TIMEOUT}s" -count=1 -cover -coverprofile coverage.out $(go list ./...)
+GO111MODULE=on go test -race -v -p=1 -timeout="${TEST_TIMEOUT}s" -count=1 -cover -coverprofile coverage.out $(go list ./...)
 go tool cover -html coverage.out -o coverage.html
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -42,10 +42,34 @@ type (
 )
 
 type testEnv struct {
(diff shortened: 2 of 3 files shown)


maruina:
Thanks @phisco, it looks great.

I think that here we have the benefit of avoiding changing the public API, but as we discussed - and in the previous PR as well - we're using the context to propagate objects between different steps. I think we should come up with a different way to expose that to the user.

phisco:
See https://github.com/kubernetes-sigs/e2e-framework/compare/main...phisco:e2e-framework:fix/tParallel-passing-almost-everywhere?expand=1 for another take at this, preserving the ctx in the env only to allow sharing information from the setup to the features and the finish phase, but avoiding the continuous thread safe update of the ctx of this PR, by passing it down the whole chain similarly to what was done by @maruina in #265, but also returning it to allow inspecting the content afterward if needed

phisco:
Closing this as I think #292 is better, assuming we want to keep the context as a way to pass information between steps. If that's not the case, which would be a major breaking change in the framework's design, but would definitely make sense, the original PR #265 from @maruina is definitely the way to go.

PR #297, opened 2023-07-13
fix: CreateWithConfig handling empty image

### LINKED ISSUE #296: error when calling `envfuncs.CreateKindClusterWithConfig` with an empty image
Kind would be fine if called with an empty image (`kind create cluster --image ""`), as long as it's an empty string, resulting in the default image being used. But given that the `image` is passed as is [here](https://github.com/kubernetes-sigs/e2e-framework/blob/612165028ab40fa2e821050c8fdb2868769a27d5/support/kind/kind.go#L98), to be able to create a Kind cluster with a custom config, but the default image, we have to use a dirty hack and pass it two empty double quotes as follows:
```golang
envfuncs.CreateKindClusterWithConfig(clusterName, "\"\"", kindCfg)
```
I think `cluster.CreateWithConfig` could definitely either quote the image before passing it as an argument to `--image` or just avoid adding the flag if the image provided is an empty string.

I'd be fine with any of my suggestions above, but if erroring out on the missing image was intentional, I'd be fine also adding a dedicated function allowing to create a kind cluster with a custom config but the default image, let me know what you would prefer.

phisco:
Fixes #296.

### DIFF
--- a/support/kind/kind.go
+++ b/support/kind/kind.go
@@ -95,7 +95,11 @@ func (k *Cluster) clusterExists(name string) (string, bool) {
 }
 
 func (k *Cluster) CreateWithConfig(imageName, kindConfigFile string) (string, error) {
-	return k.Create("--image", imageName, "--config", kindConfigFile)
+	args := []string{"--config", kindConfigFile}
+	if imageName != "" {
+		args = append(args, "--image", imageName)
+	}
+	return k.Create(args...)
 }
 
 func (k *Cluster) Create(args ...string) (string, error) {


maruina (approved):
LGTM

support/kind/kind.go
```diff
@@ -95,7 +95,7 @@ func (k *Cluster) clusterExists(name string) (string, bool) {
 }
 
 func (k *Cluster) CreateWithConfig(imageName, kindConfigFile string) (string, error) {
-	return k.Create("--image", imageName, "--config", kindConfigFile)
+	return k.Create("--image", fmt.Sprintf("%q", imageName), "--config", kindConfigFile)
```
cpanato, line 98: maybe if the image is empty we remove the `--image` flag from the create instead of setting an empty string?

wdyt @vladimirvivien

phisco: Totally agree, and I would avoid quoting it then, so our hack of using `”\”\””` won’t even break. Just let me know if you want me to switch to:
```golang

func (k *Cluster) CreateWithConfig(imageName, kindConfigFile string) (string, error) {
	args := []string{"--config", kindConfigFile}
	if imageName != "" {
		args = append(args, "--image", imageName)
	}
	return k.Create(args...)
}

```

harshanarayana: +1 to what @cpanato In case if the image is an empty string, let us just drop the argument. Makes it more readable.

phisco: I agree, updated the PR accordingly 🙏

Issue #289, opened 2023-06-27
panic handler hides source of panic


chlunde:
If there is a panic in a env.Setup EnvFunc, the source line is logged as env.go. Printing the stack trace or top frame would be helpful.

https://github.com/kubernetes-sigs/e2e-framework/blob/15e91144243aab9c611f5706c3e7c48ceb881850/pkg/env/env.go#L345-L354

Result on panic:
```
E0626 23:32:53.995230   25448 env.go:353] Recovering from panic and running finish actionsreflect: call of reflect.Value.Type on zero Value
```

harshanarayana:
I think it would be nice to dump the entire stack itself. 

`klog.ErrorS(e, "Recovering from panic and running finish actions", "stack", string(debug.Stack()))`

I believe something like this could be done. I noticed the formatting of the string was a bit off too because of how we were using `klog.Error` instead of `klog.Errorf` or `klog.ErrorS`

@chlunde  Would you be interested in opening a PR for this ?

PR #299, opened 2023-07-18
show stacktrace when panic in env.Setup

reetasingh:
fixes https://github.com/kubernetes-sigs/e2e-framework/issues/289 by adding stacktrace

### DIFF
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"fmt"
 	"regexp"
+	"runtime/debug"
 	"sync"
 	"testing"
 
@@ -346,7 +347,7 @@ func (e *testEnv) Run(m *testing.M) int {
 			if e.cfg.DisableGracefulTeardown() {
 				panic(rErr)
 			}
-			klog.Error("Recovering from panic and running finish actions", rErr)
+			klog.Errorf("Recovering from panic and running finish actions", rErr, string(debug.Stack()))
 		}
 
 		finishes := e.getFinishActions()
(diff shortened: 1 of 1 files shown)


pkg/env/env.go
```diff
@@ -346,7 +347,7 @@ func (e *testEnv) Run(m *testing.M) int {
 			if e.cfg.DisableGracefulTeardown() {
 				panic(rErr)
 			}
-			klog.Error("Recovering from panic and running finish actions", rErr)
+			klog.Errorf("Recovering from panic and running finish actions", rErr, string(debug.Stack()))
```
harshanarayana, line 350: If you are using `klog.Errorf` you would have to use the `%` format specifiers. Otherwise it will just get appended without space. 

```suggestion
			klog.ErrorS(rErr, "Recovering from panic and running finish actions", "stack", string(debug.Stack()))
```
might be a better way to setup the logger ?

harshanarayana (changes requested):

pkg/env/env.go
reetasingh: `klog.ErrorS(rErr, "Recovering from panic and running finish actions", "stack", string(debug.Stack()))` did not work because rErr is not of error type. 
I fixed previous by adding % format specifiers

pkg/env/env.go
```diff
@@ -346,7 +347,7 @@ func (e *testEnv) Run(m *testing.M) int {
 			if e.cfg.DisableGracefulTeardown() {
 				panic(rErr)
 			}
-			klog.Error("Recovering from panic and running finish actions", rErr)
+			klog.Errorf("Recovering from panic and running finish actions: %s, stack: %s", rErr, string(debug.Stack()))
```
harshanarayana, line 350: Should this be `%v` instead?

PR #303, opened 2023-08-04
Add flux HelmRepository and HelmRelease support with example

matrus2:
/kind feature

Hello again,
This time I would like to extend flux functionality and add support for HelmRepository and HelmRelease objects. There is a simple kyverno example of how to use these objects.

Docs provided in README file.


matrus2:
I see this one is still pending, can I politely request for the review @vladimirvivien in spare time. Thx

vladimirvivien:
@matrus2 apologies. Been away and distracted with work.
Taking a look today.

PR #322, opened 2023-09-13
bugfix MutateAnnotations

Smeb:
#### What type of PR is this?
/kind bug

Previously, if an object has no annotations, new annotations would be created and set as the label, and then overrides would be set on the labels.

```release-note
fixed an issue with MutateAnnotations when the object had no existing annotations
```

### DIFF
--- a/klient/decoder/decoder.go
+++ b/klient/decoder/decoder.go
@@ -254,7 +254,7 @@ func MutateAnnotations(overrides map[string]string) DecodeOption {
 		annotations := obj.GetAnnotations()
 		if annotations == nil {
 			annotations = make(map[string]string)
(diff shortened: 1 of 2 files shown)


harshanarayana (commented):
/lgtm

cc @vladimirvivien  @ShwethaKumbla

harshanarayana:
`bugfixes an issue with MutateAnnotations when the object had no existing annotations` -> Mind changing this a slight bit to read `fixed an issue with MutateAnnotations when the object had no existing annotations` ? @Smeb

vladimirvivien (approved):
Thanks @Smeb 

/lgtm

PR #345, opened 2023-11-01
Add --cluster-provider-option flag

cartermckinnon:
#### What type of PR is this?

/kind feature
/kind api-change

This adds a `--cluster-provider-option` flag that functions the same as `--labels`, to be used to pass arbitrary options to a test's `support.E2EClusterProvider`.

I think swapping `flags.FlagMap` in for `flags.LabelsMap` is safe, because while it's a public type, the `envconf` functions return a bare map (and so don't change). If there's a concern about breakage, I can try to alias the type.

```release-note
Add `--cluster-provider-option` flag for passing arbitrary options to a test's cluster provider.
```

NONE

### DIFF
--- a/pkg/envconf/config.go
+++ b/pkg/envconf/config.go
@@ -36,15 +36,16 @@ type Config struct {
 	namespace               string
 	assessmentRegex         *regexp.Regexp
 	featureRegex            *regexp.Regexp
-	labels                  flags.LabelsMap
+	labels                  flags.FlagMap
 	skipFeatureRegex        *regexp.Regexp
-	skipLabels              flags.LabelsMap
+	skipLabels              flags.FlagMap
 	skipAssessmentRegex     *regexp.Regexp
 	parallelTests           bool
 	dryRun                  bool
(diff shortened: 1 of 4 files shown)


harshanarayana:
@cartermckinnon Thanks for the contribution. This looks like a very useful thing to have. Would you mind adding an example of what this use-case would look like to make this a bit more clear for consumers ?

cartermckinnon:
Thanks for taking a look @harshanarayana, I'll add an example!

I'm at EKS, and am using this project to test our node AMI: https://github.com/awslabs/amazon-eks-ami

One use-case I had in mind is launching nodes of different EC2 instance type, or a configurable API server version.

ShwethaKumbla (commented):
@cartermckinnon Thanks for your PR.

LGTM

cartermckinnon:
I don't plan to move forward with this PR -- I've instead been using kubetest2 deployers to create the test environment, and calling the test cases written using this framework using the kubetest2 `exec` tester.

/close

PR #348, opened 2023-11-04
support/kind: create kubeconfig with stdout

### LINKED ISSUE #347: Failed to create cluster with kind podman
### What happened?

Error returned in `CreateCluster` function.

Error:
```
F1104 16:46:26.675423   12330 env.go:375] Setup failure: error loading config file "/var/folders/gk/jhz_72bs7830pb6pq1ywhcm00000gp/T/kind-cluser-test-58ebc2614499a2d22e72edbf54f-kubecfg1049500896": yaml: line 2: mapping values are not allowed in this context
```

Content of `/var/folders/gk/jhz_72bs7830pb6pq1ywhcm00000gp/T/kind-cluser-test-58ebc2614499a2d22e72edbf54f-kubecfg1049500896`:

```
enabling experimental podman provider
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUMvakNDQWVhZ0F3SUJBZ0lCQURBTkJna3Foa2lHOXcwQkFRc0ZBREFWTVJNd0VRWURWUVFERXdwcmRXSmwKY201bGRHVnpNQjRYRFRJek1URXdOREE0TkRZeE0xb1hEVE16TVRFd01UQTRORFl4TTFvd0ZURVRNQkVHQTFVRQpBeE1LYTNWaVpYSnVaWFJsY3pDQ0FTSXdEUVlKS29aSWh2Y05BUUVCQlFBRGdnRVBBRENDQVFvQ2dnRUJBS3ZXCnAzdURIUVIvKzlTUGlpeVpQTU4wSHVTSUg5QUc1eEUvZGxnbmpmSkJweXVKa0pYanE5bGVSVTF4Qmw0QmJEbWwKT2NaZUVTZ3N4blEyaU1KNEl4QSt0MWNaNk01bmVsQlRsdU8zM2krYmNaQks3UEh6TVJBL0p1V0lWTnV4S3FuRgpJWmpqY0dWeFBrT3MwS2JVcTJIbWpRUWpuZE41eWkzZkJsazRpWGNmdWU0Y2psS25uQmYwNjBYL0tZNG9oek5ICi94REdCekFvSzNiU3pzZWpPa0VRZTZIZVM5Z1Z1a3Y0UXVjTFdzY2h1dkR2UXYzclNrbGZZQmVNN3ZwZE5aRisKaEhGUzhuZVIwM2dmNy9rNzhMYysxQm1kVndkc2paQ3JUMEY4bzhwY01WempQSUFyQkhIQU1XS3BqVCtnaTlpago5YlRkK21xS3Avd3FLc01sRWxjQ0F3RUFBYU5aTUZjd0RnWURWUjBQQVFIL0JBUURBZ0trTUE4R0ExVWRFd0VCCi93UUZNQU1CQWY4d0hRWURWUjBPQkJZRUZObXkwMk8xVWRaeEdkVFVPQWhFMUwyMkFOSENNQlVHQTFVZEVRUU8KTUF5Q0NtdDFZbVZ5Ym1WMFpYTXdEUVlKS29aSWh2Y05BUUVMQlFBRGdnRUJBR3dXTDU5ZGtJZEkvYUFEM25iTwovL05aZUJGUHVvNS9OTjNsWjZ5SGp6U2hRT2RnNEM3aDI4Z29XUHMxRk50Z3ZNN1Q2U0l2TUd1b05wdUY3RmRUClJ3R1NEU2N1OFVja0lXdUJ3cjI1VHhzaVdqVGs5a2l0QVp1dzgyN21pd2Z0OGdiMGFJRU94S3ZrTkx4Y3ZqRGUKTHYvRFNVTXVWazdRSnNSV2hYdGZObUFCSXZBcEt6YXd0NDVicllPNFpSVGtBSVRCNGVjNGFoZ3BnZ1RpYzIvKwpJNmVESlRBVGpFbDNtMHcwS1V4STFMS0xwY1BwYk1pSytyNE5uT2tVN055Mmk4NjdneVBqT1llUEtXVExXR2NNCmtTS2ZWbmc5L3p0QzFQbGpHQktuNTVrbDlXMUtiRVAxaEZyR3ArMlUrTGRXMEdwWW95bEJTNmZGaEdjK1A4VVEKd0xZPQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg==
    server: https://127.0.0.1:49978
  name: kind-test-58ebc2614499a2d22e72edbf54f
contexts:
- context:
    cluster: kind-test-58ebc2614499a2d22e72edbf54f
    user: kind-test-58ebc2614499a2d22e72edbf54f
  name: kind-test-58ebc2614499a2d22e72edbf54f
current-context: kind-test-58ebc2614499a2d22e72edbf54f
kind: Config
preferences: {}
users:
- name: kind-test-58ebc2614499a2d22e72edbf54f
  user:
    client-certificate-data: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSURJVENDQWdtZ0F3SUJBZ0lJUDdZMW92eXJpdUl3RFFZSktvWklodmNOQVFFTEJRQXdGVEVUTUJFR0ExVUUKQXhNS2EzVmlaWEp1WlhSbGN6QWVGdzB5TXpFeE1EUXdPRFEyTVROYUZ3MHlOREV4TURNd09EUTJNVFJhTURReApGekFWQmdOVkJBb1REbk41YzNSbGJUcHRZWE4wWlhKek1Sa3dGd1lEVlFRREV4QnJkV0psY201bGRHVnpMV0ZrCmJXbHVNSUlCSWpBTkJna3Foa2lHOXcwQkFRRUZBQU9DQVE4QU1JSUJDZ0tDQVFFQXBRMzVXT3BmaEJObXpoV20KOUNFSERKOHdrL3QyNlZHbitXZmRZRDRETmZIZXN5VEtkMDMrSDhyT2FRUUlLVXdpYitmQ1l2QmhGUEJwaE1VeAorNjVlY2dDdTZFVG4wZndDanhJS2hCSnZiYlBDTUtENlp1OXhOdy9BZEljaG1zbGI4VVh6aG0vY1Nkc3A1OWppCk40cUZkNzA0VUxzMEV3UVMvdnNlQk4xZ2tVU2NGSm5IMUt1R284eGRieFp3Nnc2N21GMVlnSndoOE5ka25vOTEKUnltd0h1eThYakt2NnFJRHYxWlh2Z1drSkF4TElQT3kzcG5jV29uTldBV0ttUlVrckNWb1U4UVkxRWszZjdaTgpVaWxjSEcyTzFuWUVMdHZjRFlXQUMyTkVPc0FGVkVybURoT2xXVXR6akdNcTFEdzBUbXFMNGptbEQyWEw2UmpkCjRpdFRUUUlEQVFBQm8xWXdWREFPQmdOVkhROEJBZjhFQkFNQ0JhQXdFd1lEVlIwbEJBd3dDZ1lJS3dZQkJRVUgKQXdJd0RBWURWUjBUQVFIL0JBSXdBREFmQmdOVkhTTUVHREFXZ0JUWnN0Tmp0VkhXY1JuVTFEZ0lSTlM5dGdEUgp3akFOQmdrcWhraUc5dzBCQVFzRkFBT0NBUUVBZE9BL0RmTzNyczYxRTdWWWY1NVJRR3VQemRXaVBIb0tEOFFjCnc1MllNVVQ1T1ZnNXptTmV4cHNoZm5Oc2loSjNhUnN6MkFrMVVSTnpoemJjRkhGMXUzT05VVm8vcUFHTVF2MDYKcEVOek1QWWUxVG5IblZFWG1BUkRSMTlTektQeUpOekRMc3o4cW5Yd3VBU1ZreFJIUDY0dU0xVWRMbVRKc1B0dQo0ejRYNWNpZnFJR0ROMlpvT1lJVFNHbzRYU2FtWjFXMUNuQzl1R3BmSlBJUUs2TTJjRFBtR1FHSXlDcUdYZkNICjBtWkk2R09hTXNteUFTandxZU9NMGJ3dUhrcXQ1Q09Ia2xPeHNrWGVLcjhWM0laODNGdFJkNGErNGtKdVJMOEIKczE5MnZYRjFIQUd2UUhic1U5Ky9VcU5jSDRrNUZ1VlNJdjI5ZldMV28rcFdyeHdjSlE9PQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg==
    client-key-data: LS0tLS1CRUdJTiBSU0EgUFJJVkFURSBLRVktLS0tLQpNSUlFb3dJQkFBS0NBUUVBcFEzNVdPcGZoQk5temhXbTlDRUhESjh3ay90MjZWR24rV2ZkWUQ0RE5mSGVzeVRLCmQwMytIOHJPYVFRSUtVd2liK2ZDWXZCaEZQQnBoTVV4KzY1ZWNnQ3U2RVRuMGZ3Q2p4SUtoQkp2YmJQQ01LRDYKWnU5eE53L0FkSWNobXNsYjhVWHpobS9jU2RzcDU5amlONHFGZDcwNFVMczBFd1FTL3ZzZUJOMWdrVVNjRkpuSAoxS3VHbzh4ZGJ4Wnc2dzY3bUYxWWdKd2g4TmRrbm85MVJ5bXdIdXk4WGpLdjZxSUR2MVpYdmdXa0pBeExJUE95CjNwbmNXb25OV0FXS21SVWtyQ1ZvVThRWTFFazNmN1pOVWlsY0hHMk8xbllFTHR2Y0RZV0FDMk5FT3NBRlZFcm0KRGhPbFdVdHpqR01xMUR3MFRtcUw0am1sRDJYTDZSamQ0aXRUVFFJREFRQUJBb0lCQUdWVG05VGJKVUlteXdFUApSOHd0am05UU52bFI4ZUV3a09kS3ZSdTlWaWxDMEEyNE0xZGttUEtaaFhmdzByWFBtSUNNamNoQmU4Zm9BMDlQCjQreDNVbUc0UzZncHpRc2JJOU1ESjFPSEljTCtET1BmbjNtYjJ5d0xUb0g3QjBWSUFscXZMOTVEUThCUGZIMXAKcHc2anltWUUwdEswNmpZV1ZUY1ZvSXNxQUlPTTEzNXRIemlUMHpHWUYrNW9weG1MQkpMbGRWcSszVzBzbmhNbwpZTnVKcnpNYVIyWGxGcWIwZEZTdDd2UFFURC9MTmR3RzY2cXJhRkdIa2QyZ3Y3bjhSb1BBa3Y1TkV0NWFHWGRnCjk4Y2ticW5ibVhCUm9RZXl1OEQwT3lMdTlhR3dXYUJ3OEJGMVNrSCtmbFNXVlpIK3VCaTlYT1g2MnBiTHVmUUwKZTBvOTFpRUNnWUVBMTNOK0E1V24vbFI5Q0ZzNW05TCs1djR5eVJFeGFsRnpyaEk0SlVEMFBzV2NQYnVNVjBmRQpubjhZYms4RGxsVGZKQk5ScWsyWUZCSE5mYUVtaTlGY1YyN3dBV1Nla0Zmb2ZYMk5Va3Ryb3pSSC83N0x5R0gxCitlU096QUN2ZHFkVFBsZlc2T2FoeGo4NTF0Tm1rUVNyWjNRS2RnOVNGWkRzbFJOMjBpNHU4Z2tDZ1lFQXhCNWIKNDF6aGtZMXZqcWl4d0lpNnhzQXFYaWpMRWZOYXdrT0lrV2RwY1QzNFNEZnFNTlJTd0c1VHNDczNPT1JqTWZiVQpYMktMV1Y5SHVpYjdhbzBIdDlDZVFJM0JUWmdiazBzVk9xRXBiM0JDenJkSS85V0FqcUxBR0pNODFPVSsreklEClAwVFIrUVFwckx6VFlYY0VaU1AvZVFVVDdwakJ1L0xXQ3ArRm1DVUNnWUFCMHhXUEhEWlcvZ290QnZDKzlCTjYKaTA2TXkxdDZ4Z1BQeHBHNUp3cXFlV3Z3VGVWTFcrMDdWbUl6bFBHMzFUSXFrUXRSQU9VM3prcktOSklwL3ZUawpQS2xGcWI5NkEzU1JrMzJkNDh1ZWp6UEk0Nm5BSWF3K21JUHBFRnhzQ2RDM3AzREhlaVNHbEVSK0lENHZIallCClkwY3Zac1BRTVBELzlqRThhRzIzV1FLQmdRREJ5QTJiVm9jN3ZyamdDU2U1eW1CalNqV1MzRlBxNUw1ay9UM1EKRHBKTlpxSVJHYzY1ZU5KYmVhZ1BobVFHUHdVVXRZT2F6RkJNZXF6dzd0dTFxS3hib0RMK0xoTTJqSG1TekxtcApPTjBzVTR4dm5RTHdaeEdqM29SWXhBTnppU0QySzdDeHIxcVo1dGNXNjIwaC9PdkcxQTcya1dIa2UrUFJ4bHhnCkdzZHlVUUtCZ0YxL3FHSFhqcGpOZm1TdmRZTmNJMVY3VXVyVWNBbmxGOUZ4b0owallwWlFLZjhXUjF3NTg5YkYKakdqV0JWN1lQYW9wOHg4TXhqVHNGKzRiaXppOXJ3bHM4SmVrVGRHRHh5WU81Mk5jdktTZ0tMVmJsbVV0TldYTQpFbXY3MUM3M29hc29IRVp2aUVQSk11cXlBS2FUc2dvR3BBOWtUWHpNVXlQTkxES1lZN01nCi0tLS0tRU5EIFJTQSBQUklWQVRFIEtFWS0tLS0tCg==
```

I believe there is a bug in this line(https://github.com/kubernetes-sigs/e2e-framework/blob/main/support/kind/kind.go#L119) where `p.Out()` is actually outputting both stdout and stderr(https://github.com/vladimirvivien/gexe/blob/main/exec/proc.go#L206), causing the kubeconfig content to be incorrect.

Able to create a kind podman cluster.

Code to reproduce:

```
package e2e

import (
	"os"
	"testing"

	"sigs.k8s.io/e2e-framework/pkg/env"
	"sigs.k8s.io/e2e-framework/pkg/envconf"
	"sigs.k8s.io/e2e-framework/pkg/envfuncs"
	"sigs.k8s.io/e2e-framework/support/kind"
)

var testenv env.Environment

func TestMain(m *testing.M) {
	testenv, _ = env.NewFromFlags()
	kindClusterName := envconf.RandomName("test", 32)

	testenv.Setup(
		envfuncs.CreateCluster(kind.NewProvider(), kindClusterName),
	)

	os.Exit(testenv.Run(m))
}
```

Version:

kind: 0.20.0
podman: 4.6.2

### E2E Provider Used

kind

### `e2e-framework` Version

v0.3.0

### OS version

MacOS M2 chip

heylongdacoder:
#### What type of PR is this?
/kind bug

The current kubeconfig is created with both stdout and stderr. We should only use stdout to create the kubeconfig.

Fixes #347 

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/support/kind/kind.go
+++ b/support/kind/kind.go
@@ -110,17 +110,14 @@ func (k *Cluster) WithOpts(opts ...support.ClusterOpts) support.E2EClusterProvid
 func (k *Cluster) getKubeconfig() (string, error) {
 	kubecfg := fmt.Sprintf("%s-kubecfg", k.name)
 
-	p := utils.RunCommand(fmt.Sprintf(`%s get kubeconfig --name %s`, k.path, k.name))
-	if p.Err() != nil {
-		return "", fmt.Errorf("kind get kubeconfig: %w", p.Err())
+	var stdout, stderr bytes.Buffer
+	err := utils.RunCommandWithSeperatedOutput(fmt.Sprintf(`%s get kubeconfig --name %s`, k.path, k.name), &stdout, &stderr)
+	if err != nil {
+		return "", fmt.Errorf("kind get kubeconfig: stderr: %s: %w", stderr.String(), err)
 	}
+	log.V(4).Info("kind get kubeconfig stderr \n", stderr.String())
 
-	var stdout bytes.Buffer
-	if _, err := stdout.ReadFrom(p.Out()); err != nil {
-		return "", fmt.Errorf("kind kubeconfig stdout bytes: %w", err)
-	}
-
-	file, err := os.CreateTemp("", fmt.Sprintf("kind-cluser-%s", kubecfg))
+	file, err := os.CreateTemp("", fmt.Sprintf("kind-cluster-%s", kubecfg))
 	if err != nil {
 		return "", fmt.Errorf("kind kubeconfig file: %w", err)
 	}
@@ -128,7 +125,7 @@ func (k *Cluster) getKubeconfig() (string, error) {
 
 	k.kubecfgFile = file.Name()
 
-	if n, err := io.Copy(file, &stdout); n == 0 || err != nil {
+	if n, err := io.WriteString(file, stdout.String()); n == 0 || err != nil {
 		return "", fmt.Errorf("kind kubecfg file: bytes copied: %d: %w]", n, err)
 	}
 
--- a/support/utils/command.go
+++ b/support/utils/command.go
@@ -18,6 +18,7 @@ package utils
 
 import (
 	"fmt"
+	"io"
 
 	"github.com/vladimirvivien/gexe"
 	"github.com/vladimirvivien/gexe/exec"
@@ -75,10 +76,23 @@ func FindOrInstallGoBasedProvider(pPath, provider, module, version string) (stri
 	return "", fmt.Errorf("%s not available even after installation", provider)
 }
 
+// RunCommand run command and returns an *exec.Proc with information about the executed process.
 func RunCommand(command string) *exec.Proc {
 	return commandRunner.RunProc(command)
 }
 
+// RunCommandWithSeperatedOutput run command and returns the results to the provided
+// stdout and stderr io.Writer.
+func RunCommandWithSeperatedOutput(command string, stdout, stderr io.Writer) error {
+	p := commandRunner.NewProc(command)
+	p.SetStdout(stdout)
+	p.SetStderr(stderr)
+	result := p.Run()
+
+	return result.Err()
+}
+
+// FetchCommandOutput run command and returns the combined stderr and stdout output.
 func FetchCommandOutput(command string) string {
 	return commandRunner.Run(command)
 }


tingsl409:
N
Definitely, nice to have!

heylongdacoder:
seems like the `TestKwokCluster` is completely broken, I can try to have a look. But should I open another Issue and PR for that? Or I can do it with this PR? 😄

heylongdacoder:
The test passed. 😄

harshanarayana:
@heylongdacoder  You are right. That test has been flaking a bit lately. I will take a look at that over the weekend to see if we can fix that once and for all. 

/lgtm 

/cc @vladimirvivien

ShwethaKumbla (commented):
@heylongdacoder Thanks for your PR.

LGTM

support/kind/kind.go
```diff
@@ -110,7 +110,7 @@ func (k *Cluster) WithOpts(opts ...support.ClusterOpts) support.E2EClusterProvid
 func (k *Cluster) getKubeconfig() (string, error) {
 	kubecfg := fmt.Sprintf("%s-kubecfg", k.name)
 
-	p := utils.RunCommand(fmt.Sprintf(`%s get kubeconfig --name %s`, k.path, k.name))
+	p := utils.RunCommand(fmt.Sprintf(`%s get kubeconfig --name %s -q`, k.path, k.name))
```
vladimirvivien, line 113: @heylongdacoder my only concern here is that if there are other problems with retrieving the kubeconfig, that command will also stay quiet. This can be bad downstream for a running test, causing it to fail in mysterious ways since that command appears to be working when in fact, there are errors.

vladimirvivien (commented):
Hi @heylongdacoder this is a good idea, but I have some concern (see comment). My suggestion is to add an option to force the user to make the decision of running this in quiet mode. That way the developer can indicate that they understand the implication and are Ok with it.

heylongdacoder:
Thanks for the review, everyone!

@vladimirvivien, I agree with your suggestion. Just want to confirm before I implement. Is that your idea is to add one more field to `kind.Cluster`(https://github.com/kubernetes-sigs/e2e-framework/blob/main/support/kind/kind.go#L43) like `getKubeConfigOpts []string`? 😄

vladimirvivien:
@heylongdacoder yes, you will need a field to capture that configuration 

```go
*Cluster.WithQuiet()
```
Or something like that.

vladimirvivien:
@heylongdacoder originally, yes that was the intention. When/if someone uses `Cluster.WithQuiet`, it would append the `-q` to the command and therefore ignores errors. But, as you pointed out, it would 

With respect to your new suggestion, how would separating the output help ?  Let me know.

Another approach is to add a new method for getting the kubeconfig with options.

```go
func (k *Cluster) GetKubeconfigWithOption(opts ...support.ClusterOpts) (string, error) {
...
}
```

Let me know if that makes sense.

heylongdacoder:
> With respect to your new suggestion, how would separating the output help ? Let me know.

Screenshot 2023-11-20 at 2 25 34 PM

^This is the combined stderr and stdout output.

Screenshot 2023-11-20 at 2 25 49 PM

^This is stdout output

Screenshot 2023-11-20 at 2 41 22 PM

^This is stderr output

@vladimirvivien This is the result I obtained with `kind v0.17.0`. Therefore, using only stdout to create the kubeconfig can also resolve this issue. 😄

vladimirvivien:
@heylongdacoder
 I get a different result running on MacOS

```
 kind get kubeconfig --name kind
apiVersion: v1
clusters:
...
```
Notice that there is no extraneous text in the combined about, just the config.
Is this behavior tied to your environment ?

I am asking these questions to make sure this will be a one-off and that it is a general fix/enhancement.

heylongdacoder:
@vladimirvivien I guess this only happen when you are using `podman` as `kind` provider. I also found another merged PR(https://github.com/kubernetes-sigs/e2e-framework/pull/84) that fixed this issue. 😄

vladimirvivien:
@heylongdacoder Ok. So can this PR be closed ? Or does #84 fixes your issue?

heylongdacoder:
@vladimirvivien I think there was a regression. In #84, the kubeconfig was being retrieved from stdout (https://github.com/kubernetes-sigs/e2e-framework/pull/84/files#diff-c66ce9879f2c64ba698cc9f0688530c2068a838797e98b94080045eb1122d235R58). However, in the latest code (https://github.com/kubernetes-sigs/e2e-framework/blob/main/support/kind/kind.go#L119), it seems to be using both stderr and stdout again. 😄

I think the fixes in #84 make sense, I would like to a similar fix. 🙏

vladimirvivien:
Ok @heylongdacoder , can this PR put back what got regressed in #84 and we'll (re)start from there.

support/utils/command.go
```diff
@@ -82,3 +83,14 @@ func RunCommand(command string) *exec.Proc {
 func FetchCommandOutput(command string) string {
 	return commandRunner.Run(command)
 }
+
+func FetchCommandStdoutStderr(command string) (stdout, stderr string, err error) {
```
vladimirvivien, line 87: Hi @heylongdacoder couple of suggestions:

1. The "Fetch" prefix does not really describe that the command is "executed" (I see there is precedance, so maybe Ok)
2. Name suggestion: "RunWithSeparatedOutput" ?

Also, whichever name you select, please add a comment over the function to describe its operation.

vladimirvivien (commented):
Thanks for this.

support/utils/command.go
heylongdacoder: @vladimirvivien agreed. Since we have "RunCommand", I renamed it to "RunCommandWithSeperatedOutput", is this better or function name too long? 😆 

I also added some comment on the functions. 😄

support/utils/command.go
```diff
@@ -75,10 +76,24 @@ func FindOrInstallGoBasedProvider(pPath, provider, module, version string) (stri
 	return "", fmt.Errorf("%s not available even after installation", provider)
 }
 
+// RunCommand run command and returns an *exec.Proc with information about the executed process.
 func RunCommand(command string) *exec.Proc {
 	return commandRunner.RunProc(command)
 }
 
+// RunCommandWithSeperatedOutput run command and returns the result as stdout, stderr and err.
+func RunCommandWithSeperatedOutput(command string) (stdout, stderr string, err error) {
```
vladimirvivien, line 85: @heylongdacoder I thought about this change further.
I have a suggestion. Since this method is splitting the  command output streams, maybe it should accept the stream target as io streams instead of returning strings like so:

```go
func RunCommandWithSeperatedOutput(command string, stdout, stderr io.Writer) (err error) {
}
```
That way, the streams can be processed as needed (i.e. piped to a file, byte buffers, etc)

Cajga:
We run into the same. Thanks @heylongdacoder for the PR. Will you have time to finish the PR?

heylongdacoder:
hey @Cajga, recently got very busy with my new job. Probably I will have time to continue this weekend. Thanks!

vladimirvivien:
@heylongdacoder Any progress on this front ?

vladimirvivien (commented):
@heylongdacoder this looks good.
It should be tested on both podman and kind.

cc @harshanarayana 

/lgtm

heylongdacoder:
yes, tested with podman and kind on my M1 chip Macbook. 😄

harshanarayana (approved):
/lgtm
/approve

PR #349, opened 2023-11-07
support/kwok: Print process output due to an error

wzshiming:
#### What type of PR is this?

/kind feature

Fixes #

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/support/kwok/kwok.go
+++ b/support/kwok/kwok.go
@@ -139,7 +139,11 @@ func (k *Cluster) Create(ctx context.Context, args ...string) (string, error) {
 	klog.V(4).Info("Launching:", command)
 	p := utils.RunCommand(command)
 	if p.Err() != nil {
-		return "", fmt.Errorf("failed to create kwok cluster: %s : %s", p.Err(), p.Result())
+		outBytes, err := io.ReadAll(p.Out())
+		if err != nil {
+			klog.ErrorS(err, "failed to read data from the kwok create process output due to an error")
+		}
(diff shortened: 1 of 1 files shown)


support/kwok/kwok.go
```diff
@@ -139,7 +139,13 @@ func (k *Cluster) Create(ctx context.Context, args ...string) (string, error) {
 	klog.V(4).Info("Launching:", command)
 	p := utils.RunCommand(command)
 	if p.Err() != nil {
-		return "", fmt.Errorf("failed to create kwok cluster: %s : %s", p.Err(), p.Result())
+		var data []byte
```
vladimirvivien, line 142: I thought `p.Result()` does the copying of the buffers from both stderr and stdout ?

vladimirvivien (commented):
I think this can be simplified with p.Result().

support/kwok/kwok.go
wzshiming: https://github.com/kubernetes-sigs/e2e-framework/blob/b4b509b58edc6aafd85069e67397766a8967152d/support/kind/kind.go#L174-L181

This is what I copied from, `p.Result()` didn't output stdout when having error.

https://github.com/vladimirvivien/gexe/blob/35a3a027afb0ad069d6f22154df739134c921052/exec/proc.go#L229-L232

vladimirvivien: Still, can we do something like this to clean it up?

```go
str, _ := bufio.NewReader(p.Out()).ReadString(0)
```

wzshiming: done

vladimirvivien (approved):
/lgtm

PR #350, opened 2023-11-07
Allow imports of `types` package

cartermckinnon:
#### What type of PR is this?

/kind feature
/kind api-change

This moves the `types` package out of the `internal` path.

The interfaces defined in `types` cannot be imported if they are `internal`, which prevents sensible uses like `[]types.Feature` in a test func.

```release-note
NONE
```

```docs

```


vladimirvivien (commented):
Hi @cartermckinnon I think we can make this simpler by using the alias types as is done here.

https://github.com/kubernetes-sigs/e2e-framework/blob/365c4a4aadf3f71e09e4a123cfa53a590ac82a11/pkg/env/env.go#L36

We can start there. Let me know if that works.

cartermckinnon:
@vladimirvivien thanks, didn't see that!

I could add `types.Feature`, etc. to this list, but if we're going to selectively export some of these interfaces/types anyway, we could do that by moving them to a non-internal package.

vladimirvivien:
@cartermckinnon early on when the code was evolving, there was a need to shield the API since it was in flux. At this point, the API is stable(ish). Moving it to a visible place should not have an impact since it was hidden prior. I cant think of a down side doing this.

@harshanarayana what do you think?

cartermckinnon:
@vladimirvivien that makes sense, thanks!

Do you think I should mark the exports of these types in `env` as deprecated? Looks like there will be a decent amount of find-and-replace necessary for things like `env.Func`, but that can come later.

vladimirvivien:
@cartermckinnon The types in package `env` are aliases to the types in types.go.  They are designed to provide a more ergonomic access to the type values (i.e. `env.Func` for environment functions instead of the more generic, but equivalent, name `types.EnvFunc`). I think they can say, but be documented as being aliases.

harshanarayana:
+1 for this from me. I think we have no reason to keep these private.

cartermckinnon:
Hi folks, cleaning out my PR dashboard -- are there any other things that need to be addressed here?

cpanato (approved):
/lgtm

/assign @harshanarayana @vladimirvivien

PR #362, opened 2023-12-08
After setup error, do klog.Error not klog.Fatal, and move on

### LINKED ISSUE #188: Gracefully exit when setup fail
https://github.com/kubernetes-sigs/e2e-framework/blob/d19e8b569fb2f94568f8ed74b00227f879798aa0/pkg/env/env.go#L369

Due to the usage of `klog.Fatalf()`, it calls internally `os.Exit()`, and we can't gracefully run the clean ups.

codegold79:
When an error happens during a setup step, instead of calling `klog.Fatal` which causes `os.Exit`, instead do `klog.Error`, `break`, run `defer` function with clean up steps, and `return`.

#### What type of PR is this?
/kind bug

I found [this line](https://github.com/kubernetes-sigs/e2e-framework/blob/d19e8b569fb2f94568f8ed74b00227f879798aa0/pkg/env/env.go#L369) to be the problem as to why our KinD cluster deletion was not happening as intended. As soon as an error value is passed during a setup step, `klog.Fatalf(...)` is called:

```go
for _, setup := range setups {
    // context passed down to each setup
    if e.ctx, err = setup.run(e.ctx, e.cfg); err != nil {
        klog.Fatalf("%s failure: %s", setup.role, err)
   }
}
```

According to [klog docs](https://github.com/kubernetes/klog/blob/6ded8085b27eeaf68a13d583a382b125ef1482ee/klog.go#L1606-L1607),

> // Fatal logs to the FATAL, ERROR, WARNING, and INFO logs,
> // prints stack trace(s), then calls OsExit(255).

To reiterate, `klog.Fatal` is stopping all operations and exiting program with `os.Exit`. This is not the behavior I want. What I want to happen is for the setup range loop to stop trying to do the setup steps, and try to get to the function return, `return m.Run()`. However, before returning, I expect the defer block to run next where the finishing steps, like cluster cleanup, happen.

Fixes #188 

```release-note
Gracefully exit when setup fails
```

```docs

```

### DIFF
--- a/examples/graceful_setup_fail/main_test.go
+++ b/examples/graceful_setup_fail/main_test.go
@@ -1,0 +1,126 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package setupfail
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/e2e-framework/pkg/env"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+)
+
+func TestMain(m *testing.M) {
+	var testEnv env.Environment
+
+	tests := []struct {
+		name             string
+		testEnvGenerator func(context.Context, *[]string) env.Environment
+		expected         []string
+	}{
+		{
+			name: "No test setup failures, do all setup and finish actions",
+			testEnvGenerator: func(ctx context.Context, actions *[]string) env.Environment {
+				testEnv = env.New()
+
+				testEnv.Setup(
+					func(ctx context.Context, _ *envconf.Config) (context.Context, error) {
+						*actions = append(*actions, "completed setup 1")
+						return ctx, nil
+					},
+					func(ctx context.Context, _ *envconf.Config) (context.Context, error) {
+						*actions = append(*actions, "completed setup 2")
+						return ctx, nil
+					},
+					func(ctx context.Context, _ *envconf.Config) (context.Context, error) {
+						*actions = append(*actions, "completed setup 3")
+						return ctx, nil
+					},
+				)
+				testEnv.Finish(
+					func(ctx context.Context, _ *envconf.Config) (context.Context, error) {
+						*actions = append(*actions, "completed finish 1")
+						return ctx, nil
+					},
+					func(ctx context.Context, _ *envconf.Config) (context.Context, error) {
+						*actions = append(*actions, "completed finish 2")
+						return ctx, nil
+					},
+				)
+
+				return testEnv
+			},
+			expected: []string{"completed setup 1", "completed setup 2", "completed setup 3", "completed finish 1", "completed finish 2"},
+		},
+		{
+			name: "Skip remaining setup actions after one fails, do finish actions",
+			testEnvGenerator: func(ctx context.Context, actions *[]string) env.Environment {
+				testEnv = env.New()
+
+				testEnv.Setup(
+					func(ctx context.Context, _ *envconf.Config) (context.Context, error) {
+						*actions = append(*actions, "completed setup 1")
+						return ctx, nil
+					},
+					func(ctx context.Context, _ *envconf.Config) (context.Context, error) {
+						*actions = append(*actions, "completed setup 2")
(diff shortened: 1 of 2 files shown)


codegold79:
I'm thinking about how to write a test that ensures finish/clean up steps happen after setup step fails. This PR will be a WIP until I can get some tests up.

neolit123 (commented):
/ok-to-test

examples/graceful_setup_fail/main_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/e2e-framework/pkg/env"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+var testEnv env.Environment
+
+func TestMain(m *testing.M) {
```
vladimirvivien, line 33: Hi @codegold79 
TestMain is usually reserved for setting up test resources, not running tests itself.

You probably can make your test simpler by moving your test environment into a proper TestXXX function.
There is nothing special about the testEnv variable, you can declare within a test and add step to it with no problem.

Check out these tests here - https://github.com/kubernetes-sigs/e2e-framework/blob/main/pkg/env/env_test.go

examples/graceful_setup_fail/main_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"strings"
+	"testing"
+
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/e2e-framework/pkg/env"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+)
+
+var testEnv env.Environment
```
vladimirvivien, line 31: Also, if you are going to test the construction of testEnv, you may not want it to be global. Just create a new one as needed.

vladimirvivien (commented):
HI @codegold79 thanks for taking on this. This is a good start, I left some comments about the structure of your test.

examples/graceful_setup_fail/main_test.go
codegold79: Thanks @vladimirvivien for taking the time to review and provide feedback.
> TestMain is usually reserved for setting up test resources, not running tests itself...
> You probably can make your test simpler by moving your test environment into a proper TestXXX function.

I might be missing or misunderstanding something, but I wasn't able to figure out how to put the tests into a TestXXX function. At first, I tried to add a test to `func TestEnv_APIMethods(t *testing.T)` at https://github.com/kubernetes-sigs/e2e-framework/blob/332536f9ed7ed22fd069c5622b057a91ceeb1208/pkg/env/env_test.go#L46C6-L46C24. I also tried to add a test to `func TestEnv_Test(t *testing.T)` at https://github.com/kubernetes-sigs/e2e-framework/blob/332536f9ed7ed22fd069c5622b057a91ceeb1208/pkg/env/env_test.go#L137C1-L137C32.

With both TestXXX functions, I had the same problem: the [defer block](https://github.com/kubernetes-sigs/e2e-framework/blob/332536f9ed7ed22fd069c5622b057a91ceeb1208/pkg/env/env.go#L353-L377) is not reached in `func (e *testEnv) Run(m *testing.M) (exitCode int)` at https://github.com/kubernetes-sigs/e2e-framework/blob/332536f9ed7ed22fd069c5622b057a91ceeb1208/pkg/env/env.go#L345C1-L345C52.

I believe to reach that block, my only option is to call `func (e *testEnv) Run(m *testing.M) (exitCode int)`. But that funciton requires passing in `m *testing.M`. And the only way I know to pass that variable in is via `TestMain(...)`. Hench why I added an example package to do the testing, as all the examples use TestMain(...).

The reason why I believe that[defer block](https://github.com/kubernetes-sigs/e2e-framework/blob/332536f9ed7ed22fd069c5622b057a91ceeb1208/pkg/env/env.go#L353-L377) needs to be called is that's where the finish steps are run in case an error happens during a setup step.

I hope despite my lack of brevity, I was able to show my limited understanding, and perhaps you or another person might help me figure out a better way to do the testing.

examples/graceful_setup_fail/main_test.go
codegold79: Will do!

vladimirvivien:
@codegold79 apologies for the delay.
let me know if this ready for another round of review.

codegold79:
Thanks, @vladimirvivien! I responded to your first review comment and tried to explain why I had to make a separate example and use `TestMain`. 

As for your second review comment, I made the change as you suggested. I also got rid of the dummy test. as it doesn't help anything except to not trigger a warning that there are no tests. The thing is, I'm not testing any features, I'm making sure the final actions complete even after a panic in the setup, before any feature tests happen.

PR is ready for review again.

vladimirvivien:
@codegold79 (apologies for the late reply)
I did spend sometimes to see if this can be tested using a normal test function and I think you are doing the right thing.

/lgtm

vladimirvivien:
Hi @codegold79 can you please squash commits. Thanks.

cpanato (approved):
thanks
/lgtm

PR #372, opened 2024-01-23
Resolve issue #369 by eliminating the redundant invocation of cfg.client.RestConfig in CreateClusterWithConfig, which is leading to a panic

### LINKED ISSUE #369: Ambiguous call to `cfg.Client()` in `CreateClusterWithConfig`
### What happened?

For context: I'm implementing a custom provider and I was plying around with `CreateClusterWithConfig` and `CreateCluster`. I noticed panic with `CreateClusterWithConfig` resulting from the call to `cfg.Client()` [1] which expects a kubeconfig to be in the config before being set. There is no call to `cfg.Client()` before `cg.WithKubeConfigFile()` in `CreateCluster`.

[1]: https://github.com/kubernetes-sigs/e2e-framework/blob/main/pkg/envfuncs/provider_funcs.go#L84

The call before initialization looks like a sideeffect. Not sure if it is wanted.

---

_No response_

### E2E Provider Used

kind

### `e2e-framework` Version

0.3.0

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

reetasingh:
#### What type of PR is this?

/kind bug 
/kind regression

Remove unnecessary call to `cfg.client.RestConfig` in `CreateClusterWithConfig` . The cfg.Client() function expects cfg.kubeconfig to be set before calling. This is raising panic. Also the cfg.Client().RESTConfig() returns a restconfig which is not used anywhere in that function. 

Fixes # https://github.com/kubernetes-sigs/e2e-framework/issues/369

```release-note
```

```docs

```

### DIFF
--- a/pkg/envfuncs/provider_funcs.go
+++ b/pkg/envfuncs/provider_funcs.go
@@ -81,7 +81,6 @@ func CreateClusterWithConfig(p support.E2EClusterProvider, clusterName, configFi
 			return ctx, err
 		}
 
-		cfg.Client().RESTConfig()
 		// update envconfig  with kubeconfig
 		cfg.WithKubeconfigFile(kubecfg)
 


harshanarayana:
fixes #369 369

PR #375, opened 2024-02-06
use longer time if deadline is set

sivchari:
#### What type of PR is this?

/kind feature

If the context deadline is set and the time is longer than default wait time, deadline time is preferred.

Fixes #

```release-note
```

```docs

```

### DIFF
--- a/support/kwok/kwok.go
+++ b/support/kwok/kwok.go
@@ -138,7 +138,7 @@ func (k *Cluster) Create(ctx context.Context, args ...string) (string, error) {
 	waitTime, ok := ctx.Deadline()
 	// If the deadline is set and the wait time is less than the time left before the deadline, then we should use the wait time
 	if ok && waitTime.Before(time.Now().Add(wait)) {
-		wait = waitTime.Sub(time.Now())
+		wait = time.Until(waitTime)
 	}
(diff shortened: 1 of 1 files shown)


vladimirvivien (commented):
Hi @sivchari 
Since kwok is a cluster provider, should this behavior implemented with other cluster providers like Kind?
cc @harshanarayana

harshanarayana:
+1 to what @vladimirvivien  mentioned. Guess this can be a general thing across each providers ?

sivchari:
Kind has --wait flag, but it's not used in cluster provider. If we'll use it, we should new parameter to wait. Should I change it ?

vladimirvivien:
@sivchari (apologies for late reply)
How is the `Duration` value passed in to Kwok as a provider ?
We should use the same mechanism to pass it to kind if possible.
If a flag is needed, we can explore that.

sivchari:
KWOK has already had `waitDruration` in struct. So if Kind also has wautDuration in [Cluster](https://github.com/kubernetes-sigs/e2e-framework/blob/main/support/kind/kind.go#L43-L50), we can control the duration time when to wait, can't it ?

vladimirvivien:
@sivchari yes, but I am not sure if that is already built into kind (which would make it easier to implement).

Issue #288, opened 2023-06-26
WithStep level argument, defined `Levels` are private


phisco:
As per the title, I noticed the [WithStep](https://github.com/kubernetes-sigs/e2e-framework/blob/15e91144243aab9c611f5706c3e7c48ceb881850/pkg/features/builder.go#L42) accepts a `Level` as argument, but the actual levels are only declared in an internal package and not publicly exposed, https://github.com/kubernetes-sigs/e2e-framework/blob/15e91144243aab9c611f5706c3e7c48ceb881850/pkg/internal/types/types.go#L105, is this intentional?

I would have liked to use `WithStep` to distinguish `Assessment`s with side-effects from Assessments only doing actual assessments. Not really a blocker, obviously.

harshanarayana:
@phisco  Sorry for taking a bit to respond back to this. Entirely missed responding back to this. 

I think it should be ok to expose the Level values outside of the internal package alone. @vladimirvivien  do you have any concerns on doing that ?

phisco:
No problem, wasn't really a blocker for us in the end, just a nice to have.

vladimirvivien:
@phisco Missed the original post. Thanks for pointing that out. 
Originally, levels were meant as an internal mechanism to distinguish the different steps.
Now that more folks are using the framework, if it's useful to access that info at runtime, I am not against that.

phisco:
Awesome! I'll open a PR later today 👍

PR #376, opened 2024-02-07
refactor: expose Levels

phisco:
#### What type of PR is this?

/kind cleanup

[WithStep](https://github.com/kubernetes-sigs/e2e-framework/blob/15e91144243aab9c611f5706c3e7c48ceb881850/pkg/features/builder.go#L42) accepts a Level as argument, but the actual levels (Setup, Assess, Teardown)  are only declared in an internal package and not publicly exposed. This PR exposes the internal specific Levels.

Fixes https://github.com/kubernetes-sigs/e2e-framework/issues/288

Has to be said that any other Level w.r.t. the 3 intended ones will be ignored, as the code explicitly selects the three steps, so we could also go the other way, remove the public Level and avoid exposing WithStep and WithDescribeStep private, but that would be a breaking change unfortunately.

```release-note
```

```docs

```

### DIFF
--- a/pkg/features/feature.go
+++ b/pkg/features/feature.go
@@ -30,6 +30,15 @@ type (
 	Level   = types.Level
 )
 
+const (
+	// LevelSetup when doing the setup phase
+	LevelSetup Level = types.LevelSetup
+	// LevelAssess when doing the assess phase
+	LevelAssess = types.LevelAssess
+	// LevelTeardown when doing the teardown phase
+	LevelTeardown = types.LevelTeardown
+)
+
 type defaultFeature struct {
 	name        string
 	description string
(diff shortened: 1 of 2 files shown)


pkg/features/feature.go
```diff
@@ -30,6 +30,15 @@ type (
 	Level   = types.Level
 )
 
+const (
+	// LevelSetup when doing the setup phase
+	LevelSetup Level = types.LevelSetup
```
harshanarayana, line 35:
```suggestion
	LevelSetup = types.LevelSetup
```
This should be good enough ?

harshanarayana (changes requested):
Just a minor comment. Looks good otherwise.

harshanarayana (approved):
/lgtm
/approve

PR #381, opened 2024-02-15
Fix DeleteNamespace not retrieving ns from ctx

Fricounet:
#### What type of PR is this?

/kind bug

The function DeleteNamespace was not succeeding in retrieving the namespace from the context because it tried to cast a corev1.Namespace to *corev1.Namespace.

Fixes #

```release-note
```

```docs

```

### DIFF
--- a/pkg/envfuncs/ns_funcs.go
+++ b/pkg/envfuncs/ns_funcs.go
@@ -60,8 +60,8 @@ func DeleteNamespace(name string) env.Func {
 		// attempt to retrieve from context
 		nsVal := ctx.Value(namespaceContextKey(name))
 		if nsVal != nil {
-			if ns, ok := nsVal.(*corev1.Namespace); ok {
-				namespace = ns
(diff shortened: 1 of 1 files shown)


cpanato (commented):
/lgtm

/assign @harshanarayana

vladimirvivien (commented):
Again, thanks @Fricounet .
Is there a test for that delete case?  If not, can you add one?  Thank you.

Fricounet:
@vladimirvivien there were no tests for the delete. If that is fine with you (and because it will be a mess to refactor the 2 PRs otherwise), I've created a [test](https://github.com/kubernetes-sigs/e2e-framework/pull/382/commits/9f3b76f9d5168d8161d38e9200ec91966637245c) for `DeleteNamespace` in the other PR I opened (#382). Is that ok for you?

vladimirvivien:
Yes that is ok. I just want to use opportunity to add a test to avoid a future bug. Thanks again .

PR #382, opened 2024-02-15
Allow custom labels and annotations for namespace creation

### LINKED ISSUE #364: Allow adding custom labels and annotations when creating namespace
### What do you want to see?

Hello :wave: 
I was wondering if we could consider allowing to add user-defined labels and annotation when creating a namespace with [envfuncs.CreateNamespace](https://github.com/kubernetes-sigs/e2e-framework/blob/v0.3.0/pkg/envfuncs/ns_funcs.go#L38).

The feature request is motivated by the fact that in our case we use custom labels and annotations to configure behaviors of the namespace like cascade deletion of the namespace even if resources still exist in it, or adding metadata related to which feature is running in the namespace.
What do you think about it?

For the implementation, I think it could be easily added with an option pattern:
```
type CreateNamespaceOpts func(ns *corev1.Namespace)

func WithLabels(client klient.Client, labels map[string]string) CreateNamespaceOpts {
  return func(ns *corev1.Namespace) {
    client.Ressources.Labels(&ns, labels)
  }
}

func WithAnnotations(client klient.Client, annotations map[string]string) CreateNamespaceOpts {
	return func(ns *corev1.Namespace) {
		client.Ressources.Annotate(&ns, annotations)
	}
}

func CreateNamespace(name string, opts ...CreateNamespaceOpts) env.Func {
	return func(ctx context.Context, cfg *envconf.Config) (context.Context, error) {
		namespace := corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: name}}
		client, err := cfg.NewClient()
		if err != nil {
			return ctx, fmt.Errorf("create namespace func: %w", err)
		}
               for _, opt := range opts {
		        opt(client, &namespace)
	       }
		if err := client.Resources().Create(ctx, &namespace); err != nil {
			return ctx, fmt.Errorf("create namespace func: %w", err)
		}
		cfg.WithNamespace(name) // set env config default namespace
		return context.WithValue(ctx, namespaceContextKey(name), namespace), nil
	}
}
```

### Extra Labels

_No response_

Fricounet:
#### What type of PR is this?

/kind feature

Create 2 new option functions that can be used when calling `CreateNamespace` to add custom labels and annotations to the namespace.

Fixes #364 

I also added tests to make sure the function work as expected (this is how I discovered #381), however it looks like there is no automated testing in place. In the meantime, tests can be run with `go test ./pkg/envfuncs`.

To be merged after #381 

```release-note
Added a way to create namespaces with custom labels and annotations
```

```docs

```

### DIFF
--- a/pkg/envfuncs/main_test.go
+++ b/pkg/envfuncs/main_test.go
@@ -21,11 +21,22 @@ import (
 	"testing"
 
 	"sigs.k8s.io/e2e-framework/pkg/env"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/envfuncs"
+	"sigs.k8s.io/e2e-framework/support/kind"
 )
 
-var testenv env.Environment
+// nsTestenv is used to test the ns_funcs
+var nsTestenv env.Environment
 
 func TestMain(m *testing.M) {
-	testenv = env.New()
-	os.Exit(testenv.Run(m))
+	nsClusterName := envconf.RandomName("ns-cluster", 16)
+	nsTestenv = env.New()
+
+	// Use the same cluster for all ns_funcs tests
+	nsTestenv.
+		Setup(envfuncs.CreateCluster(kind.NewProvider(), nsClusterName)).
+		Finish(envfuncs.DestroyCluster(nsClusterName))
+
+	os.Exit(nsTestenv.Run(m))
 }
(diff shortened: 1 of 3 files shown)


cpanato (commented):
/ok-to-test

vladimirvivien (commented):
Thanks @Fricounet
This looks great to me.
@harshanarayana PTAL, if you are OK, please approve.

harshanarayana (approved):
/lgtm
/approve

PR #395, opened 2024-04-02
fix(support): propagate kind image options to CLI

jashandeep-sohi:
#### What type of PR is this?

/kind bug

The current support E2EClusterProvider API for creating Kind clusters is confusing/misleading.
Say, I want to create a Kind cluster with a custom image:

```go
import "sigs.k8s.io/e2e-framework/support/kind"

p := kind.NewProvider().WithOpts(kind.WithImage("..."))
```

When used with `envfuncs.CreateCluster` I would expect this to work, but when you dig down to it, the image name is never passed down to the CLI arguments:
https://github.com/kubernetes-sigs/e2e-framework/blob/fdcf509f96c96f0055b03f026ea44c62df82c753/support/kind/kind.go#L170

Confusingly, the other variant `kind.CreateWithConfig` does https://github.com/kubernetes-sigs/e2e-framework/blob/fdcf509f96c96f0055b03f026ea44c62df82c753/support/kind/kind.go#L153

The behavior should be the same for both.

### DIFF
--- a/support/kind/kind.go
+++ b/support/kind/kind.go
@@ -150,9 +150,6 @@ func (k *Cluster) CreateWithConfig(ctx context.Context, kindConfigFile string) (
 	if kindConfigFile != "" {
 		args = append(args, "--config", kindConfigFile)
 	}
-	if k.image != "" {
-		args = append(args, "--image", k.image)
-	}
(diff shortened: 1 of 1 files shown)


harshanarayana (commented):
/lgtm

PR #367, opened 2024-01-14
feat: make Config's Client() goroutine safe

pmalek:
#### Which issue(s) this PR fixes:

Fixes #352

#### Special notes for your reviewer:

Would we like to extend the like over all the fields of `Config` or leave it as proposed and only cover `Client()`?

#### Does this PR introduce a user-facing change?

```release-note
Config's Client() method is now goroutine safe
```

#### Additional documentation e.g., Usage docs, etc.:

```docs

```

PR #396, opened 2024-04-08
[env] Do not share config between tests

### LINKED ISSUE #352: Calling `Config`'s `Client()` in tests running parallel yields a data race
### What happened?

Data race in `pkg/envconf/config.go`

```
go test -v -race -count 1 -trimpath .
=== RUN   TestX1
=== PAUSE TestX1
=== RUN   TestX2
=== PAUSE TestX2
=== CONT  TestX1
=== CONT  TestX2
=== RUN   TestX1/feature
=== RUN   TestX2/feature
==================
WARNING: DATA RACE
Write at 0x00c000134480 by goroutine 19:
  sigs.k8s.io/e2e-framework/pkg/envconf.(*Config).Client()
      sigs.k8s.io/e2e-framework@v0.3.1-0.20231113122213-262cac32d35e/pkg/envconf/config.go:141 +0x9c
  e2eclientresources.Setup()
      e2eclientresources/main_test.go:29 +0x38
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).executeSteps()
      sigs.k8s.io/e2e-framework@v0.3.1-0.20231113122213-262cac32d35e/pkg/env/env.go:435 +0x104
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).processTestFeature.(*testEnv).execFeature.func1()
      sigs.k8s.io/e2e-framework@v0.3.1-0.20231113122213-262cac32d35e/pkg/env/env.go:449 +0x13c
  testing.tRunner()
      testing/testing.go:1595 +0x194
  testing.(*T).Run.func1()
      testing/testing.go:1648 +0x40

Previous write at 0x00c000134480 by goroutine 20:
  sigs.k8s.io/e2e-framework/pkg/envconf.(*Config).Client()
      sigs.k8s.io/e2e-framework@v0.3.1-0.20231113122213-262cac32d35e/pkg/envconf/config.go:141 +0x9c
  e2eclientresources.Setup()
      e2eclientresources/main_test.go:29 +0x38
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).executeSteps()
      sigs.k8s.io/e2e-framework@v0.3.1-0.20231113122213-262cac32d35e/pkg/env/env.go:435 +0x104
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).processTestFeature.(*testEnv).execFeature.func1()
      sigs.k8s.io/e2e-framework@v0.3.1-0.20231113122213-262cac32d35e/pkg/env/env.go:449 +0x13c
  testing.tRunner()
      testing/testing.go:1595 +0x194
  testing.(*T).Run.func1()
      testing/testing.go:1648 +0x40

Goroutine 19 (running) created at:
  testing.(*T).Run()
      testing/testing.go:1648 +0x5d8
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).execFeature()
      sigs.k8s.io/e2e-framework@v0.3.1-0.20231113122213-262cac32d35e/pkg/env/env.go:442 +0x3ac
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).processTestFeature()
      sigs.k8s.io/e2e-framework@v0.3.1-0.20231113122213-262cac32d35e/pkg/env/env.go:215 +0x234
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).processTests()
      sigs.k8s.io/e2e-framework@v0.3.1-0.20231113122213-262cac32d35e/pkg/env/env.go:277 +0x900
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).Test()
      sigs.k8s.io/e2e-framework@v0.3.1-0.20231113122213-262cac32d35e/pkg/env/env.go:326 +0x60
  e2eclientresources.TestX1()
      e2eclientresources/main_test.go:44 +0xb8
  testing.tRunner()
      testing/testing.go:1595 +0x194
  testing.(*T).Run.func1()
      testing/testing.go:1648 +0x40

Goroutine 20 (running) created at:
  testing.(*T).Run()
      testing/testing.go:1648 +0x5d8
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).execFeature()
      sigs.k8s.io/e2e-framework@v0.3.1-0.20231113122213-262cac32d35e/pkg/env/env.go:442 +0x3ac
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).processTestFeature()
      sigs.k8s.io/e2e-framework@v0.3.1-0.20231113122213-262cac32d35e/pkg/env/env.go:215 +0x234
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).processTests()
      sigs.k8s.io/e2e-framework@v0.3.1-0.20231113122213-262cac32d35e/pkg/env/env.go:277 +0x900
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).Test()
      sigs.k8s.io/e2e-framework@v0.3.1-0.20231113122213-262cac32d35e/pkg/env/env.go:326 +0x60
  e2eclientresources.TestX2()
      e2eclientresources/main_test.go:49 +0xb8
  testing.tRunner()
      testing/testing.go:1595 +0x194
  testing.(*T).Run.func1()
      testing/testing.go:1648 +0x40
==================
=== RUN   TestX1/feature/A
=== RUN   TestX2/feature/A
=== NAME  TestX1/feature
    testing.go:1465: race detected during execution of test
=== NAME  TestX2/feature
    testing.go:1465: race detected during execution of test
=== NAME  TestX1
    testing.go:1465: race detected during execution of test
--- FAIL: TestX1 (0.04s)
    --- FAIL: TestX1/feature (0.04s)
        --- PASS: TestX1/feature/A (0.00s)
=== NAME  TestX2
    testing.go:1465: race detected during execution of test
--- FAIL: TestX2 (0.04s)
    --- FAIL: TestX2/feature (0.04s)
        --- PASS: TestX2/feature/A (0.00s)
FAIL
FAIL    e2eclientresources      0.089s
FAIL
```

No data race

```
package main

import (
	"context"
	"os"
	"testing"

	corev1 "k8s.io/api/core/v1"
	"sigs.k8s.io/e2e-framework/klient/conf"
	"sigs.k8s.io/e2e-framework/pkg/env"
	"sigs.k8s.io/e2e-framework/pkg/envconf"
	"sigs.k8s.io/e2e-framework/pkg/features"
)

var tenv env.Environment

func TestMain(m *testing.M) {
	cfg, err := envconf.NewFromFlags()
	if err != nil {
		panic(err)
	}
	cfg.WithKubeconfigFile(conf.ResolveKubeConfigFile())
	tenv = env.NewWithConfig(cfg)

	os.Exit(tenv.Run(m))
}

func Setup(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
	res := c.Client().Resources()
	podList := corev1.PodList{}
	err := res.List(ctx, &podList)
	if err != nil {
		t.Fatalf("failed listing pods: %v", err)
	}
	return ctx
}

func Assess(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
	return ctx
}

func TestX1(t *testing.T) {
	t.Parallel()
	tenv.Test(t, generateFeat())
}

func TestX2(t *testing.T) {
	t.Parallel()
	tenv.Test(t, generateFeat())
}

func generateFeat() features.Feature {
	return features.New("feature").
		WithSetup("setup", Setup).
		Assess("A", Assess).
		Feature()
}
```

_No response_

### E2E Provider Used

kind

### `e2e-framework` Version

v0.3.1-0.20231113122213-262cac32d35e

### OS version

<details>

```console
$ uname -a
Darwin MBP 23.1.0 Darwin Kernel Version 23.1.0: Mon Oct  9 21:27:24 PDT 2023; root:xnu-10002.41.9~6/RELEASE_ARM64_T6000 arm64 arm Darwin
```

</details>

### LINKED ISSUE #384: Do not share config across tests
### What happened?

Currently, the `config` object is passed as a reference across all the `StepFunc` of a feature. When running tests, it creates a number of issues where any test updating the config would also change it for all other tests leading to undesirable side effects.

I currently see 2 use cases where this is an issue:
- creating a separate namespace for each test with `-parallel`. When using the `envfuncs.CreateNamespace` func, the namespace is stored in the config. However, when the tests are running in parallel, the `cfg.namespace` will be modified. There is an [example](https://github.com/kubernetes-sigs/e2e-framework/tree/main/examples/every_test_custom_ns) going around the issue using the context however I feel like it is hacky compared to using the cfg object
- changing the config for a specific test. For instance, if I want to enable `FailFast` on a specifc test (because I know that when it fails there is no point continuing it), I can't enable it for this specific test because when I do `cfc.WithFailFast()`, it is gonna apply to all the remaining tests running after.

This code is enough to show the issue:
```
var testenv env.Environment

func TestMain(m *testing.M) {
	clusterName := envconf.RandomName("cluster", 16)
	testenv = env.New()

	testenv.
		Setup(envfuncs.CreateCluster(kind.NewProvider(), clusterName)).
		BeforeEachTest(
			func(ctx context.Context, cfg *envconf.Config, t *testing.T) (context.Context, error) {
				t.Parallel()
				namespace := envconf.RandomName("ns", 16)
				return envfuncs.CreateNamespace(namespace)(ctx, cfg)
			},
		).
		Finish(envfuncs.DestroyCluster(clusterName))

	os.Exit(testenv.Run(m))
}

func Test1(t *testing.T) {
	feat := features.New("Test1").Assess("Test1", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
		time.Sleep(1 * time.Second) // sleep to wait for test 2
		t.Log("Namespace 1: ", cfg.Namespace())
		return ctx
	}).Feature()

	testenv.Test(t, feat)
}

func Test2(t *testing.T) {
	feat := features.New("Test1").Assess("Test1", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
		t.Log("Namespace 2: ", cfg.Namespace())
		return ctx
	}).Feature()

	testenv.Test(t, feat)
}

```

Result: 
```
=== RUN   Test1
=== PAUSE Test1
=== RUN   Test2
=== PAUSE Test2
=== CONT  Test1
=== CONT  Test2
=== RUN   Test2/Test1
=== RUN   Test1/Test1
=== RUN   Test2/Test1/Test1
=== RUN   Test1/Test1/Test1
=== NAME  Test2/Test1/Test1
    main_test.go:64: Namespace 2:  ns-6ef56304fde1a
--- PASS: Test2 (0.02s)
    --- PASS: Test2/Test1 (0.00s)
        --- PASS: Test2/Test1/Test1 (0.00s)
=== NAME  Test1/Test1/Test1
    main_test.go:55: Namespace 1:  ns-6ef56304fde1a
--- PASS: Test1 (1.02s)
    --- PASS: Test1/Test1 (1.00s)
        --- PASS: Test1/Test1/Test1 (1.00s)
PASS
```

I think what we could do would be to make copies of the config at different steps of the tests (before each test, before each feature, before each assess) and only share the copy with the child element (test config -> feature config -> assess config). This way, each step could override the cfg without any risk of impacting the parent or cousins (what happens in the assess would not impact what happens in the other assess or the feature). However, modifying the feature config could impact all the assess in the feature.
What do you think about this?

### E2E Provider Used

real cluster

### `e2e-framework` Version

v0.3.0

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

Fricounet:
#### What type of PR is this?

/kind feature

Following discussions in https://github.com/kubernetes-sigs/e2e-framework/pull/367

This PR fixes a bug where the config object was shared between tests. When running tests in parallel, it was then impossible to rely on fields like the namespace because they could have been overwritten by another test. Also, it led to tests using `-race` to fail because the shared config.klient object that were updating the same fields when initializing the client.

This PR creates a new `deepCopyConfig` method that allows to create a deep copy of the config object. This way, each test can have its own without impacting the others. Now the config has the following lifecycle:
- It is created when a testEnv is created
- Each test uses a child testEnv that inherits the main testEnv's config
- Each feature uses a child testEnv that inherits the test's testEnv's config

This way, a feature inherits all the changes made in BeforeEach functions while not impacting the other features.

Fixes #352
Fixes #384 

```release-note
Improve how Config is shared across tests so that it can be used without any issues in parallel tests
```

```docs

```

### DIFF
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -29,6 +29,7 @@ import (
 
 	klog "k8s.io/klog/v2"
 
+	"sigs.k8s.io/e2e-framework/klient"
 	"sigs.k8s.io/e2e-framework/pkg/envconf"
 	"sigs.k8s.io/e2e-framework/pkg/featuregate"
 	"sigs.k8s.io/e2e-framework/pkg/features"
@@ -113,6 +114,18 @@ func newTestEnvWithParallel() *testEnv {
 	return &testEnv{
 		ctx: context.Background(),
 		cfg: envconf.New().WithParallelTestEnabled(),
+	}
+}
+
+// newChildTestEnv returns a child testEnv based on the one passed as an argument.
+// The child env inherits the context and actions from the parent and
+// creates a deep copy of the config so that it can be mutated without
+// affecting the parent's.
+func newChildTestEnv(e *testEnv) *testEnv {
+	return &testEnv{
+		ctx:     e.ctx,
+		cfg:     e.deepCopyConfig(),
+		actions: e.actions,
 	}
 }
 
@@ -241,7 +254,8 @@ func (e *testEnv) processFeatureActions(ctx context.Context, t *testing.T, featu
 // In case if the parallel run of test features are enabled, this function will invoke the processTestFeature
 // as a go-routine to get them to run in parallel
 func (e *testEnv) processTests(ctx context.Context, t *testing.T, enableParallelRun bool, testFeatures ...types.Feature) context.Context {
-	if e.cfg.DryRunMode() {
+	dedicatedTestEnv := newChildTestEnv(e)
+	if dedicatedTestEnv.cfg.DryRunMode() {
 		klog.V(2).Info("e2e-framework is being run in dry-run mode. This will skip all the before/after step functions configured around your test assessments and features")
 	}
 	if ctx == nil {
@@ -251,19 +265,20 @@ func (e *testEnv) processTests(ctx context.Context, t *testing.T, enableParallel
 		t.Log("No test testFeatures provided, skipping test")
 		return ctx
 	}
-	beforeTestActions := e.getBeforeTestActions()
-	afterTestActions := e.getAfterTestActions()
+	beforeTestActions := dedicatedTestEnv.getBeforeTestActions()
+	afterTestActions := dedicatedTestEnv.getAfterTestActions()
 
-	runInParallel := e.cfg.ParallelTestEnabled() && enableParallelRun
+	runInParallel := dedicatedTestEnv.cfg.ParallelTestEnabled() && enableParallelRun
 
 	if runInParallel {
 		klog.V(4).Info("Running test features in parallel")
 	}
 
-	ctx = e.processTestActions(ctx, t, beforeTestActions)
+	ctx = dedicatedTestEnv.processTestActions(ctx, t, beforeTestActions)
 
 	var wg sync.WaitGroup
 	for i, feature := range testFeatures {
+		featureTestEnv := newChildTestEnv(dedicatedTestEnv)
 		featureCopy := feature
 		featName := feature.Name()
 		if featName == "" {
@@ -273,13 +288,13 @@ func (e *testEnv) processTests(ctx context.Context, t *testing.T, enableParallel
 			wg.Add(1)
 			go func(ctx context.Context, w *sync.WaitGroup, featName string, f types.Feature) {
 				defer w.Done()
-				_ = e.processTestFeature(ctx, t, featName, f)
+				_ = featureTestEnv.processTestFeature(ctx, t, featName, f)
 			}(ctx, &wg, featName, featureCopy)
 		} else {
-			ctx = e.processTestFeature(ctx, t, featName, featureCopy)
+			ctx = featureTestEnv.processTestFeature(ctx, t, featName, featureCopy)
 			// In case if the feature under test has failed, skip reset of the features
 			// that are part of the same test
-			if e.cfg.FailFast() && t.Failed() {
+			if featureTestEnv.cfg.FailFast() && t.Failed() {
 				break
 			}
 		}
@@ -287,7 +302,7 @@ func (e *testEnv) processTests(ctx context.Context, t *testing.T, enableParallel
 	if runInParallel {
 		wg.Wait()
 	}
-	return e.processTestActions(ctx, t, afterTestActions)
+	return dedicatedTestEnv.processTestActions(ctx, t, afterTestActions)
 }
 
 // TestInParallel executes a series a feature tests from within a
@@ -572,7 +587,49 @@ func (e *testEnv) requireProcessing(kind, testName string, requiredRegexp, skipR
 	return skip, message
 }
 
-// deepCopyFeature just copies the values from the Feature but creates a deep
+// deepCopyConfig just copies the values from the Config to create a deep
+// copy to avoid mutation when we just want an informational copy.
+func (e *testEnv) deepCopyConfig() *envconf.Config {
+	// Basic copy which takes care of all the basic types (str, bool...)
+	configCopy := *e.cfg
+
+	// Manually setting fields that are struct types
+	if e.cfg.KubeconfigFile() != "" { // No kubeconfig means than no client could have been created
+		clientCopy, _ := klient.New(e.cfg.Client().RESTConfig())
+		configCopy.WithClient(clientCopy)
+	}
+	if e.cfg.AssessmentRegex() != nil {
+		configCopy.WithAssessmentRegex(e.cfg.AssessmentRegex().String())
+	}
+	if e.cfg.FeatureRegex() != nil {
+		configCopy.WithFeatureRegex(e.cfg.FeatureRegex().String())
+	}
+	if e.cfg.SkipAssessmentRegex() != nil {
+		configCopy.WithSkipAssessmentRegex(e.cfg.SkipAssessmentRegex().String())
+	}
+	if e.cfg.SkipFeatureRegex() != nil {
+		configCopy.WithSkipFeatureRegex(e.cfg.SkipFeatureRegex().String())
+	}
+
+	labels := make(map[string][]string, len(e.cfg.Labels()))
+	for k, vals := range e.cfg.Labels() {
+		copyVals := make([]string, len(vals))
+		copyVals = append(copyVals, vals...)
+		labels[k] = copyVals
+	}
+	configCopy.WithLabels(labels)
+
+	skipLabels := make(map[string][]string, len(e.cfg.SkipLabels()))
+	for k, vals := range e.cfg.SkipLabels() {
+		copyVals := make([]string, len(vals))
+		copyVals = append(copyVals, vals...)
+		skipLabels[k] = copyVals
+	}
+	configCopy.WithSkipLabels(e.cfg.SkipLabels())
+	return &configCopy
+}
+
+// deepCopyFeature just copies the values from the Feature to create a deep
 // copy to avoid mutation when we just want an informational copy.
 func deepCopyFeature(f types.Feature) types.Feature {
 	fcopy := features.New(f.Name())
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -117,15 +117,18 @@ func newTestEnvWithParallel() *testEnv {
 	}
 }
 
+type ctxName string
+
 // newChildTestEnv returns a child testEnv based on the one passed as an argument.
 // The child env inherits the context and actions from the parent and
 // creates a deep copy of the config so that it can be mutated without
 // affecting the parent's.
 func newChildTestEnv(e *testEnv) *testEnv {
+	childCtx := context.WithValue(e.ctx, ctxName("parent"), fmt.Sprintf("%s", e.ctx))
 	return &testEnv{
-		ctx:     e.ctx,
+		ctx:     childCtx,
 		cfg:     e.deepCopyConfig(),
-		actions: e.actions,
+		actions: append([]action{}, e.actions...),
 	}
 }
 
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -597,8 +597,9 @@ func (e *testEnv) deepCopyConfig() *envconf.Config {
 	configCopy := *e.cfg
 
 	// Manually setting fields that are struct types
-	if e.cfg.KubeconfigFile() != "" { // No kubeconfig means than no client could have been created
-		clientCopy, _ := klient.New(e.cfg.Client().RESTConfig())
+	if client := e.cfg.GetClient(); client != nil {
+		// Need to recreate the underlying client because client.Resource is not thread safe
+		clientCopy, _ := klient.New(client.RESTConfig())
 		configCopy.WithClient(clientCopy)
 	}
 	if e.cfg.AssessmentRegex() != nil {
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -599,7 +599,11 @@ func (e *testEnv) deepCopyConfig() *envconf.Config {
 	// Manually setting fields that are struct types
 	if client := e.cfg.GetClient(); client != nil {
 		// Need to recreate the underlying client because client.Resource is not thread safe
-		clientCopy, _ := klient.New(client.RESTConfig())
+		// Panic on error because this should never happen since the client was built once already
+		clientCopy, err := klient.New(client.RESTConfig())
+		if err != nil {
+			panic(err)
+		}
 		configCopy.WithClient(clientCopy)
 	}
 	if e.cfg.AssessmentRegex() != nil {
--- a/pkg/envconf/config.go
+++ b/pkg/envconf/config.go
@@ -131,12 +131,11 @@ func (c *Config) NewClient() (klient.Client, error) {
 // are confident in the configuration or call NewClient() to ensure its
 // safe creation.
 func (c *Config) Client() klient.Client {
-	client, err := klient.NewWithKubeConfigFile(c.kubeconfig)
+	client, err := c.NewClient()
 	if err != nil {
-		panic(fmt.Errorf("envconfig: client failed: %w", err).Error())
+		panic(err)
 	}
-	c.client = client
-	return c.client
+	return client
 }
 
 // WithNamespace updates the environment namespace value
--- a/pkg/envconf/config.go
+++ b/pkg/envconf/config.go
@@ -107,6 +107,11 @@ func (c *Config) WithClient(client klient.Client) *Config {
 	return c
 }
 
+// GetClient returns the client for the environment
+func (c *Config) GetClient() klient.Client {
+	return c.client
+}
+
 // NewClient is a constructor function that returns a previously
 // created klient.Client or create a new one based on configuration
 // previously set. Will return an error if unable to do so.
--- a/pkg/envconf/config.go
+++ b/pkg/envconf/config.go
@@ -116,10 +116,6 @@ func (c *Config) GetClient() klient.Client {
 // created klient.Client or create a new one based on configuration
 // previously set. Will return an error if unable to do so.
 func (c *Config) NewClient() (klient.Client, error) {
-	if c.client != nil {
-		return c.client, nil
-	}
-
 	client, err := klient.NewWithKubeConfigFile(c.kubeconfig)
 	if err != nil {
 		return nil, fmt.Errorf("envconfig: client failed: %w", err)
@@ -135,10 +131,6 @@ func (c *Config) NewClient() (klient.Client, error) {
 // are confident in the configuration or call NewClient() to ensure its
 // safe creation.
 func (c *Config) Client() klient.Client {
-	if c.client != nil {
-		return c.client
-	}
-
 	client, err := klient.NewWithKubeConfigFile(c.kubeconfig)
 	if err != nil {
 		panic(fmt.Errorf("envconfig: client failed: %w", err).Error())
--- a/pkg/env/env_test.go
+++ b/pkg/env/env_test.go
@@ -733,7 +733,6 @@ func TestTestEnv_TestInParallel(t *testing.T) {
 var envTForChildCfgParallelTesting = NewWithConfig(envconf.New().WithParallelTestEnabled().WithNamespace("child-cfg"))
 
 func TestTestEnv_ChildCfgInParallel(t *testing.T) {
-
 	envTForChildCfgParallelTesting.BeforeEachTest(func(ctx context.Context, config *envconf.Config, t *testing.T) (context.Context, error) {
 		t.Logf("Running before each test for test %s", t.Name())
 		t.Parallel()
--- a/pkg/env/env_test.go
+++ b/pkg/env/env_test.go
@@ -726,6 +726,90 @@ func TestTestEnv_TestInParallel(t *testing.T) {
 	}
 }
 
+// Create a dedicated env that can be used to test the parallel execution of tests and features to make sure
+// they don't share the same config object but they inherit the one from the parent env.
+// Meaning that each test inherit the global testEnv and each feature inherit the testEnv of the test.
+// This env is used for the 3 tests below.
+var envTForChildCfgParallelTesting = NewWithConfig(envconf.New().WithParallelTestEnabled().WithNamespace("child-cfg"))
+
+func TestTestEnv_ChildCfgInParallel(t *testing.T) {
+
+	envTForChildCfgParallelTesting.BeforeEachTest(func(ctx context.Context, config *envconf.Config, t *testing.T) (context.Context, error) {
+		t.Logf("Running before each test for test %s", t.Name())
+		t.Parallel()
+
+		// Check that namespace was inherited from the global config
+		if config.Namespace() != "child-cfg" {
+			t.Errorf("Expected namespace to be %s but got %s", t.Name(), config.Namespace())
+		}
+		config.WithNamespace(t.Name())
+		return ctx, nil
+	})
+
+	envTForChildCfgParallelTesting.AfterEachTest(func(ctx context.Context, config *envconf.Config, t *testing.T) (context.Context, error) {
+		t.Logf("Running after each test for test %s", t.Name())
+		if config.Namespace() != t.Name() {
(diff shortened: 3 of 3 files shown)


pkg/env/env.go
```diff
@@ -116,6 +117,18 @@ func newTestEnvWithParallel() *testEnv {
...
 
+// newChildTestEnv returns a child testEnv based on the one passed as an argument.
+// The child env inherits the context and actions from the parent and
+// creates a deep copy of the config so that it can be mutated without
+// affecting the parent's.
+func newChildTestEnv(e *testEnv) *testEnv {
+	return &testEnv{
+		ctx:     e.ctx,
+		cfg:     e.deepCopyConfig(),
+		actions: e.actions,
```
phisco, line 128: Should we create a new slice here as we do in [WithContext](https://github.com/kubernetes-sigs/e2e-framework/blob/0ba3d3baf4c53aa9bc90f641a9f603ec6b295e01/pkg/env/env.go#L127), to avoid possible race conditions on that too.

Fricounet: sure, it's not a perfect copy though: since the underlying `funcs`, `featureFuncs` and `testFuncs` fields of each action are also slices, they are still shared. But I don't think there is any way in the current code to alter those slices after creation so it should not cause any issue even if we just copy the `[]action` slice.

Would you prefer a deep copy of the actions as well to be safe?

Fricounet: I've copied the `actions` like it is done in `WithContext` for now and I've also created a child context so that none of the fields of the testEnv are shared

pmalek:
I've tested this with a simple test (from #352) by checking out locally an adding a `replace`:

```
replace sigs.k8s.io/e2e-framework => ../e2e-framework
```

```
package main

import (
	"context"
	"os"
	"testing"

	corev1 "k8s.io/api/core/v1"
	"sigs.k8s.io/e2e-framework/klient/conf"
	"sigs.k8s.io/e2e-framework/pkg/env"
	"sigs.k8s.io/e2e-framework/pkg/envconf"
	"sigs.k8s.io/e2e-framework/pkg/features"
)

var tenv env.Environment

func TestMain(m *testing.M) {
	cfg, err := envconf.NewFromFlags()
	if err != nil {
		panic(err)
	}
	cfg.WithKubeconfigFile(conf.ResolveKubeConfigFile())
	tenv = env.NewWithConfig(cfg)

	os.Exit(tenv.Run(m))
}

func Setup(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
	res := c.Client().Resources()
	podList := corev1.PodList{}
	err := res.List(ctx, &podList)
	if err != nil {
		t.Fatalf("failed listing pods: %v", err)
	}
	return ctx
}

func Assess(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
	return ctx
}

func TestX1(t *testing.T) {
	t.Parallel()
	tenv.Test(t, generateFeat())
}

func TestX2(t *testing.T) {
	t.Parallel()
	tenv.Test(t, generateFeat())
}

func generateFeat() features.Feature {
	return features.New("feature").
		WithSetup("setup", Setup).
		Assess("A", Assess).
		Feature()
}
```

( I needed to add additional `replace` directives to work around #399 

and I still get a data race:

```
go test -v -race -count 1 -trimpath .
=== RUN   TestX1
=== PAUSE TestX1
=== RUN   TestX2
=== PAUSE TestX2
=== CONT  TestX2
=== CONT  TestX1
==================
WARNING: DATA RACE
Write at 0x00c00030ea80 by goroutine 26:
  sigs.k8s.io/e2e-framework/pkg/envconf.(*Config).Client()
      sigs.k8s.io/e2e-framework@v0.3.0/pkg/envconf/config.go:141 +0x9c
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).deepCopyConfig()
      sigs.k8s.io/e2e-framework@v0.3.0/pkg/env/env.go:601 +0x100
  sigs.k8s.io/e2e-framework/pkg/env.newChildTestEnv()
      sigs.k8s.io/e2e-framework@v0.3.0/pkg/env/env.go:130 +0xe4
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).processTests()
      sigs.k8s.io/e2e-framework@v0.3.0/pkg/env/env.go:260 +0x48
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).Test()
      sigs.k8s.io/e2e-framework@v0.3.0/pkg/env/env.go:346 +0x60
  e2e-framework-bug-352.TestX1()
      e2e-framework-bug-352/main_test.go:44 +0xb8
  testing.tRunner()
      testing/testing.go:1689 +0x180
  testing.(*T).Run.gowrap1()
      testing/testing.go:1742 +0x40

Previous write at 0x00c00030ea80 by goroutine 27:
  sigs.k8s.io/e2e-framework/pkg/envconf.(*Config).Client()
      sigs.k8s.io/e2e-framework@v0.3.0/pkg/envconf/config.go:141 +0x9c
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).deepCopyConfig()
      sigs.k8s.io/e2e-framework@v0.3.0/pkg/env/env.go:601 +0x100
  sigs.k8s.io/e2e-framework/pkg/env.newChildTestEnv()
      sigs.k8s.io/e2e-framework@v0.3.0/pkg/env/env.go:130 +0xe4
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).processTests()
      sigs.k8s.io/e2e-framework@v0.3.0/pkg/env/env.go:260 +0x48
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).Test()
      sigs.k8s.io/e2e-framework@v0.3.0/pkg/env/env.go:346 +0x60
  e2e-framework-bug-352.TestX2()
      e2e-framework-bug-352/main_test.go:49 +0xb8
  testing.tRunner()
      testing/testing.go:1689 +0x180
  testing.(*T).Run.gowrap1()
      testing/testing.go:1742 +0x40

Goroutine 26 (running) created at:
  testing.(*T).Run()
      testing/testing.go:1742 +0x5e4
  testing.runTests.func1()
      testing/testing.go:2161 +0x80
  testing.tRunner()
      testing/testing.go:1689 +0x180
  testing.runTests()
      testing/testing.go:2159 +0x6e0
  testing.(*M).Run()
      testing/testing.go:2027 +0xb74
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).Run()
      sigs.k8s.io/e2e-framework@v0.3.0/pkg/env/env.go:408 +0x4c0
  e2e-framework-bug-352.TestMain()
      e2e-framework-bug-352/main_test.go:25 +0x250
  main.main()
      _testmain.go:51 +0x294

Goroutine 27 (running) created at:
  testing.(*T).Run()
      testing/testing.go:1742 +0x5e4
  testing.runTests.func1()
      testing/testing.go:2161 +0x80
  testing.tRunner()
      testing/testing.go:1689 +0x180
  testing.runTests()
      testing/testing.go:2159 +0x6e0
  testing.(*M).Run()
      testing/testing.go:2027 +0xb74
  sigs.k8s.io/e2e-framework/pkg/env.(*testEnv).Run()
      sigs.k8s.io/e2e-framework@v0.3.0/pkg/env/env.go:408 +0x4c0
  e2e-framework-bug-352.TestMain()
      e2e-framework-bug-352/main_test.go:25 +0x250
  main.main()
      _testmain.go:51 +0x294
==================
=== RUN   TestX2/feature
=== RUN   TestX1/feature
=== NAME  TestX2
    testing.go:1398: race detected during execution of test
=== NAME  TestX1
    testing.go:1398: race detected during execution of test
=== RUN   TestX2/feature/A
=== RUN   TestX1/feature/A
--- FAIL: TestX2 (0.04s)
    --- PASS: TestX2/feature (0.03s)
        --- PASS: TestX2/feature/A (0.00s)
--- FAIL: TestX1 (0.04s)
    --- PASS: TestX1/feature (0.03s)
        --- PASS: TestX1/feature/A (0.00s)
FAIL
FAIL    e2e-framework-bug-352   0.532s
FAIL
```

Fricounet:
@pmalek indeed :thinking: 
So it's the client again. This stems from the fact that `Client()` function is not a proper `getter` function like the other Config accessors but a mix of a setter and a getter. As a result, when it is accessed the first time by both tests, it is `nil` and they both try to create it at the same time. I can think of a few ways of solving this:
1) Create a new client getter function like `GetClient()` which would simply acts as a getter for the client and use it in [deepCopyConfig](https://github.com/kubernetes-sigs/e2e-framework/pull/396/files#diff-688294d2a07f1d5aa5afe0db41e55097fb8de2f5f9a38537564cbad54bad4136R601) instead of `Client()`
2) Change the `Client()` function to be a proper getter and find a new name for the current `Client()`, maybe `MustNewClient()` since it panics in case of error?
3) Use a lock on the client. What you wanted to do here https://github.com/kubernetes-sigs/e2e-framework/pull/367. This would need some additional changes to make the lock available in the `env` package but I think that could work

Each has its own quirks:
1) is the easiest to setup and doesn't break the API. Drawback is that it is not inline with the current setter/getter convention on the `Config` object
2) Change the API to have saner and clearer method names. But breaks the API. However, since we are still in versions v0.X.Y, I think breaking the API can be considered fine and arguably, breaking it now to have something more future proof is OK in my opinion
3) I don't like using locks where we can avoid it especially since this lock will be exported outside of the `config` pkg to be accessible in `env`. So anybody would be able to play around with this lock while it should stay an implementation detail


For now, I went with 1) because it is the easiest to do but I'd rather do 2) if we say it is fine to break the API

phisco:
I think client caching also has another issue, so while at it, might be worth addressing that too: if one gets the client, modifies the kubeconfig or any other client related parameter in the config, and then gets the client again, they would get the original client and not a new one with the new kubeconfig. We could probably remove caching altogether and solve both issues.

pmalek:
> I think client caching also has another issue, so while at it, might be worth addressing that too: if one gets the client, modifies the kubeconfig or any other client related parameter in the config, and then gets the client again, they would get the original client and not a new one with the new kubeconfig. We could probably remove caching altogether and solve both issues.

If it's possible to go this path I'd vote for that option. 👍

Fricounet:
@phisco @pmalek I got rid of the caching by removing it altogether. 
Drawback being that the client will be recreated every time it is accessed with `config.Client()`. I tried to rename my `GetClient` function to `Client` in order to have a simple getter that won't recreate the client each time. However it creates issues where the `config.client` object is never initialized because callers of `Client` do not expect to have to initialize it themselves and we cannot initialize it ourselves within the `New...` contructor functions because it happens to early and clusters might not be available yet (see #57). My try is available [here](https://github.com/Fricounet/e2e-framework/commit/8745a4a6aacbc9b1a7eb61188c7a1c46d6db5f44) if you want to take a look.

In any case, removing the caching altogether seems like an ok solution if we are fine with the performance cost induced by having to recreate the client all the time.

pmalek:
> Drawback being that the client will be recreated every time it is accessed with `config.Client()`

I am personally not afraid of the performance costs of creating the client each time (it should be negligible, right?). Haven't measured this though.

phisco:
> > Drawback being that the client will be recreated every time it is accessed with `config.Client()`
> 
> 
> 
> I am personally not afraid of the performance costs of creating the client each time (it should be negligible, right?). Haven't measured this though.

+1, being this a testing framework I feel that shouldn't be the focus, but the call is to the maintainers to make for sure.

harshanarayana:
> I am personally not afraid of the performance costs of creating the client each time (it should be negligible, right?). Haven't measured this though.

I personally don't think this overhead in performance will be that significant if we create client each time. But we should probably keep in mind the resource usage in terms of memory. Just to account for cases where one runs these in a sandbox-ed container env. We don't want them getting OOM'ed suddenly out of nowhere right ?

Fricounet:
@harshanarayana this PR does indeed create some memory overhead. However, I did some profiling and observed the following:
- it only impacts parallel tests since serial tests will only create 1 `Config` object at a time
- each config seems to weighs around `600kB` in memory. 
  - Running 1 test in parallel:
image
  - Running 10 tests in parallel: 
image
  - Running 100 tests in parallel:
image
- this has nothing to do with caching the client or not. It is the fact that a new `Config` object has to be created for each test/feature that creates the overhead (which is the core reason for this PR in the first place).

Considering this, I think the overhead is not a big issue because it needs heavy parallelization to show a big impact. I personally don't see a situation where a system able to run 100 parallel tests would be restricted by 70MB of RAM. And considering that without the changes brought by this PR, running parallel tests with the current framework is simply not practical because of all the race conditions due to the shared config, I don't think we are losing anything here:
- serial tests don't have a noticeable overhead
- parallel tests can now be run without race conditions on the `Config`.

What do you think?



If you want to reproduce the profiling I did, I used the following code:
```
package main

import (
	"context"
	"flag"
	"fmt"
	"os"
	"runtime/pprof"
	"testing"
	"time"

	"sigs.k8s.io/e2e-framework/klient/conf"
	"sigs.k8s.io/e2e-framework/pkg/env"
	"sigs.k8s.io/e2e-framework/pkg/envconf"
	"sigs.k8s.io/e2e-framework/pkg/features"
)

var tenv env.Environment

var nParallelTests = flag.Int("n", 100, "Number of parallel tests to run")

func TestMain(m *testing.M) {
	cfg, err := envconf.NewFromFlags()
	if err != nil {
		panic(err)
	}
	cfg.WithKubeconfigFile(conf.ResolveKubeConfigFile())
	cfg.WithParallelTestEnabled()
	tenv = env.NewWithConfig(cfg)

	os.Exit(tenv.Run(m))
}

func Setup(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
	c.Client().Resources()
	time.Sleep(5 * time.Second)
	return ctx
}

func Assess(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
	return ctx
}

func TestX1(t *testing.T) {
	t.Parallel()
	feats := []features.Feature{}
	for i := 0; i < *nParallelTests; i++ {
		feats = append(feats, generateFeat())

	}
	tenv.TestInParallel(t, feats...)
}

func TestX2(t *testing.T) {
	t.Parallel()
	// Making sure all parallel features in TestX1 have stared and created their client
	time.Sleep(2 * time.Second)
	tenv.Test(t, features.New("feature").
		Assess("A", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
			file, err := os.Create(fmt.Sprintf("heap-%d.prof", *nParallelTests))
			if err != nil {
				t.Fatal(err)
			}
			defer file.Close()
			err = pprof.WriteHeapProfile(file)
			if err != nil {
				t.Fatal(err)
			}
			return ctx
		}).
		Feature())
}

func generateFeat() features.Feature {
	return features.New("feature").
		WithSetup("setup", Setup).
		Assess("A", Assess).
		Feature()
}
```

then `N=10 go test ./ -test.v -race -kubeconfig $KUBECONFIG -args -n $N && go tool pprof -web -focus env heap-$N.prof`

pmalek:
> Considering this, I think the overhead is not a big issue because it needs heavy parallelization to show a big impact. I personally don't see a situation where a system able to run 100 parallel tests would be restricted by 70MB of RAM. And considering that without the changes brought by this PR, running parallel tests with the current framework is simply not practical because of all the race conditions due to the shared config, I don't think we are losing anything here:

💯 

This can always be left as an exercise for those that would like to optimize this later on. This PR will give users the possibility to run tests in parallel without data races. Let's focus on that. If there's ever need ( I doubt that ) to make this more efficient, we can work on that then.

Thanks for all the tests that you've done @Fricounet 👍

vladimirvivien:
I appreciate the thoroughness of the analysis in this PR. Thank you all.

pkg/envconf/config.go
```diff
@@ -130,16 +131,11 @@ func (c *Config) NewClient() (klient.Client, error) {
...
-	client, err := klient.NewWithKubeConfigFile(c.kubeconfig)
+	client, err := c.NewClient()
 	if err != nil {
-		panic(fmt.Errorf("envconfig: client failed: %w", err).Error())
+		panic(err)
 	}
-	c.client = client
-	return c.client
+	return client
 }
```
phisco, line 139: ⚠️ I actually realised this would prevent injecting a client without a kubeconfig in the Config. As we would always have to go through `klient.NewWithKubeConfigFile(c.kubeconfig)` instead of relying on the injected one, which afaict is a supported usecase at the moment!

phisco: See https://github.com/kubernetes-sigs/e2e-framework/pull/413 for an attempt at solving this

Fricounet: Do you have a real example of this being an issue? 
Because currently, I can't figure out a situation where this happens. And you can still do `GetClient` to retrieve the client right?

phisco: It's just that this would be a behavior change, currently someone creating a config with an injected client would always get that same client when calling `Client()`. With this change one would have to call `GetClient()` explicitly to get that back, while calling `Client()` would result in an error if there was no kubeconfig configured for example, which is something totally acceptable if you just injected your own client.

Fricounet: ok, it makes sense then. If you can find a way to support this without overcomplicating thing, I'm all in. 
Otherwise, maybe we could consider doing a breaking change? Since we are running in version `v0.x.y`, I feel like we can afford to do breaking changes if it means we have more stable foundations to run the e2e tests at scale.

phisco: See my draft PR, unfortunately I couldn't get the tests to run yet.

PR #406, opened 2024-04-29
initialize k8s client access appropriately when a cluster exists

sanposhiho:
#### What type of PR is this?

/kind bug

ssia.
Currently, when a cluster already exists, `initKubernetesAccessClients` won't be run.
`initKubernetesAccessClients` is mandatory to be run to make `*cluster` return `KubernetesRestConfig`, and consequently now `KubernetesRestConfig` always results in `nil` in such case. 

Fixes #

```release-note
Fix a bug which both kwok and kind providers don't work expectedly when a cluster already exists.
```

```docs

```

### DIFF
--- a/support/kind/kind.go
+++ b/support/kind/kind.go
@@ -158,7 +158,11 @@ func (k *Cluster) Create(ctx context.Context, args ...string) (string, error) {
 
 	if _, ok := k.clusterExists(k.name); ok {
 		log.V(4).Info("Skipping Kind Cluster.Create: cluster already created: ", k.name)
-		return k.getKubeconfig()
+		kConfig, err := k.getKubeconfig()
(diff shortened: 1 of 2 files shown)


harshanarayana (commented):
/lgtm

harshanarayana:
@sanposhiho thanks for the PR.. Seems like the same bug exists in the `kind` provider as well . Would you mind taking care of that as well?

sanposhiho:
@harshanarayana Done.

harshanarayana:
Thanks very much @sanposhiho

harshanarayana (commented):
/lgtm 

cc @vladimirvivien @cpanato

vladimirvivien (approved):
/lgtm

PR #389, opened 2024-03-05
feat: enable feature flag to reverse test finish action

harshanarayana:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:

Changes in this PR enable the consumers of `e2e-framework` to set a feature flag that enables the execution of `Finish` actions to mimic the nature of `t.Cleanup`.

#### Which issue(s) this PR fixes:

Fixes #269

#### Special notes for your reviewer:
NA

#### Does this PR introduce a user-facing change?

```release-note
added a feature-gate argument to the supported flags that will enable setting `ReverseTestFinishExecutionOrder` so that `Finish` action can mimic `t.Cleanup`
```

#### Additional documentation e.g., Usage docs, etc.:

```docs

```

PR #413, opened 2024-05-08
Do not cache config if not explicitly set

phisco:
#### What type of PR is this?

/kind bug

Fixes https://github.com/kubernetes-sigs/e2e-framework/issues/352
Fixes https://github.com/kubernetes-sigs/e2e-framework/issues/384

Based on https://github.com/kubernetes-sigs/e2e-framework/pull/396

Based on https://github.com/kubernetes-sigs/e2e-framework/pull/396, but attempting to solve https://github.com/kubernetes-sigs/e2e-framework/pull/396#discussion_r1593786101.

This way we won't have any concurrent writes to the config client, as we are not caching it, the internal client is just for user-injected clients, so that a user can continue avoiding to provide a kubeconfig altogether if not needed. We might want to adjust naming.

```release-note
```

The client won't be cached anymore, I wouldn't call it a user-facing change.

```docs

```


phisco:
trying pinging @harshanarayana 🙏 😬

harshanarayana:
Looking into it @phisco

pkg/env/env.go
```diff
@@ -116,6 +117,21 @@ func newTestEnvWithParallel() *testEnv {
...
 }
 
+type ctxName string
+
+// newChildTestEnv returns a child testEnv based on the one passed as an argument.
+// The child env inherits the context and actions from the parent and
+// creates a deep copy of the config so that it can be mutated without
+// affecting the parent's.
+func newChildTestEnv(e *testEnv) *testEnv {
+	childCtx := context.WithValue(e.ctx, ctxName("parent"), fmt.Sprintf("%s", e.ctx))
```
harshanarayana, line 127: Why turn the parent context into a string of values here ? Since the parent context is not of much use, can this be replaced with `context.WithoutCancel(e.ctx)` instead ?

harshanarayana:
@phisco @Fricounet  Do we still need to keep https://github.com/kubernetes-sigs/e2e-framework/pull/396 alive?

phisco:
Didn't want to hijack the original pr, just wanted to see if it broke tests, I'm fine closing this one and porting only my commit to #389 🙏

pkg/env/env.go
Fricounet: The reason why I used this was because:
1) we need to create a new context from the parent's
2) we still want to propagate cancel from the parent to the child (I don't see why we would want to not do that)

Because of 2), we can't use `context.WithoutCancel`.
Which only leave `context.WithValue` and `context.WithCancel`:
- `WithCancel`: IIRC `go vet` would not let me ignore the cancel func with `_` 
- `WithValue` was not perfect because we do not need any value to be passed down but I couldn't find better

Happy to change it if you consider that we can ignore the cancellation from the parent's context though. What do you think?

Fricounet (commented):
@phisco I do not mind if we merge this one instead since it already has all the changes

pkg/env/env.go
harshanarayana: @Fricounet Good point. I see what you mean. 

The only reason I brought up this question was because we are converting the context into a string and setting it as value. Which is almost of no-use. Can we set something else as the value instead ?

harshanarayana:
/lgtm 

@vladimirvivien  @cpanato  mind taking a second look on this ?

vladimirvivien:
Thanks @phisco @harshanarayana @cpanato 

/lgtm
/approve

phisco:
And thanks @Fricounet for the initial implementation!

Issue #343, opened 2023-10-27
Expose `ko` features as a third-party Envfunc


vladimirvivien:
### What do you want to see?

[ko](https://github.com/ko-build/ko) is a tool that automates the build and publication of Go code as container images. I think it would be useful to be able to create and deploy (locally or even remotely) container images during testing. This is especially true if building/testing Kubernetes controller/operators.

For instance, the following provides an idea of how that could work :

```go
import(
    "sigs.k8s.io/e2e-framework/third_party/ko"
...
)

var testEnv env.Environment

func TestMain(m *testing.M) {
    testEnv.Setup(
        func(ctx context.Context, cfg *envconf.Config) (context.Context, error) {
            kotool = ko.Install("@latest")
            kotool.SetConfig("some-conf-file")
            kootool.RepoLogin("TOKEN_ENV")
            kotool.BuildLocal(ctx, "package-path") // builds/publishes in local docker, saves image name in ctx
            kotool.BuildRemote(ctx, "repo-path", "package-path")
            return ctx, nil
        }
    )

    testEnv.Finish(
        func(ctx context.Context, cfg *envconf.Config) (context.Context, error) {
            ko.RemoveLocalmage(ctx) // removes all built images from names saved in ctx
        }
}
```

heylongdacoder:
Hello, if no one is working on this yet, may I give it a try? 😄

vladimirvivien:
@heylongdacoder Still interested in working on this ?

/remove-lifecycle stale

heylongdacoder:
Hey @vladimirvivien, I have finished part of it with test cases. I will probably commit the code for review this weekend. But if I am blocking the progress, please feel free to take this! Many thanks!

vladimirvivien:
Hi @heylongdacoder were you able to start PR?

heylongdacoder:
@vladimirvivien sorry for the delay, I am working on it now.

vladimirvivien:
@heylongdacoder no worries at all.

heylongdacoder:
hey @vladimirvivien I would like to continue on this and confirm a few things. 

1. Are `BuildDockerLocal` and `BuildDockerRemote` still relevant? 
2. If point 1 is still relevant, for `RemoveLocalImage`, we would need the `docker` binary since `ko` does not have the functionality to remove images from Docker. Do you think that is ok?
[...]

PR #415, opened 2024-05-11
third_party/ko: Add ko as third party tools

heylongdacoder:
Add Install, BuildLocal and GetLocalImage functions.

#### What type of PR is this?
/kind feature

Already explained in #343 

Partially addresses #343 

N/A

```release-note
NONE
```

```docs
N/A
```

### DIFF
--- a/examples/third_party_integration/README.md
+++ b/examples/third_party_integration/README.md
@@ -5,3 +5,4 @@ This section of the repository contains the example of how the third party tooli
 
 1. [Helm](./helm)
 2. [Flux](./flux)
+3. [Ko](./ko)
--- a/examples/third_party_integration/ko/ko_test.go
+++ b/examples/third_party_integration/ko/ko_test.go
@@ -1,0 +1,102 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ko
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"runtime"
+	"testing"
+	"time"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"sigs.k8s.io/e2e-framework/klient/wait"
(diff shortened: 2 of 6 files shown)


heylongdacoder:
Hi @vladimirvivien, I just realized you were suggesting the BuildLocal function should build and publish to a local Docker. This PR is about ko and a KinD cluster. Since KinD is a popular way to test controllers and operators, I think this PR might be helpful. If so, maybe we can merge this PR first and then create another follow-up PR for the BuildLocal (Docker) and BuildRemote (Docker) functions. WDYT? 😄

cpanato (commented):
/ok-to-test

vladimirvivien:
@heylongdacoder this is great.
So, does this PR allow local build of images and deploy images into a running Kind cluster?

@harshanarayana cc

cpanato:
it looks great, just one question why not use `ko` as lib instead need to have that installed and use via shell?

vladimirvivien:
+1 on @cpanato question.

heylongdacoder:
> does this PR allow local build of images and deploy images into a running Kind cluster?

@vladimirvivien yes, you are right 😄 

> why not use ko as lib

@cpanato @vladimirvivien , I never think about this because other tools was installed and use via shell. But if we want to use `ko` as lib, do we need to ensure `ko` maintainers agree to maintain those functions signature?

harshanarayana:
+1 from me too.  I like the idea of integrating this as a lib instead of the CLI. 


I have been using `ko` as a lib for my work related item for a while in dev mode. They even have a doc around it. https://ko.build/advanced/go-packages/

With this in place, I suppose there is going to be some guarantee on backward compatibility.

harshanarayana:
>  So, does this PR allow local build of images and deploy images into a running Kind cluster?

We should be able to chain the build call to the kind providers load image to get this working. Right ? @vladimirvivien

tsl409:
+1 definitely nice to have!

cpanato:
I think, for now, we can keep it as is. We have other tools that follow this approach.
and as a follow up, we can migrate to use as a lib

cpanato (approved):
/approve

/hold for @harshanarayana @vladimirvivien

harshanarayana (commented):
/lgtm

vladimirvivien:
Thanks @heylongdacoder this is great.

All good points raised here. Brining ko Go dependency could be overkill and since Kind uses OS commands, I think this is Ok as well. 

As @harshanarayana mentioned, I think it a user should be able to load the docker image in Kind. Maybe an example to show the whole cycle of build and load would be useful.

heylongdacoder:
> Maybe an example to show the whole cycle of build and load would be useful.

https://github.com/kubernetes-sigs/e2e-framework/pull/415/files#diff-5e974e6375efb2beaa1644b28e57ba8e5c006b9dab1b47620b4edc24cd50d042R40, it is actually here. The KinD cluster is created in `main_test.go`. Then, the ko `BuildLocal` function is used to build the `testdata/example_goapp` and load it into the KinD cluster using its name. Do you all think this is good enough? 😄

heylongdacoder:
thanks for all the review!

PR #419, opened 2024-05-17
klient/decoder: add filename context to decode error

sttts:
#### What type of PR is this?

/kind bug

When decoding fails, the decoder has no knowledge about the filename, but `DecodeEachFile` has. The later should add it to the error context.

Fixes #

```release-note
NONE
```

```docs

```


cpanato (approved):
thanks

/lgtm

PR #421, opened 2024-06-04
Mark E2E plumbing as test helpers

negz:
#### What type of PR is this?

/kind feature

In Crossplane we try to use reusable step functions in our E2E tests, so they end up looking like this:

```go
func TestExample(t *testing.T) {
	environment.Test(t,
		features.New(t.Name()).
			WithSetup("ExampleSetup", funcs.SomeSetupFn()).
			Assess("ExampleAssess", funcs.SomeAssessFn()).
			WithTeardown("ExampleTeardown", funcs.SomeTearDownFn()).
			Feature(),
	)
}
```

Since step functions are passed `*testing.T` we'll often call `t.Error`, `t.Log`, etc from our step functions. These testing calls include the call point (i.e. filename and line number) in their output. Without using `t.Helper()` the call point will be the file where the step function is defined (e.g. `features.go`) rather than the calling test.

We want to mark our step functions as test helpers using `t.Helper` so that the calling test is included in the output. For this to work we need to mark all the intermediary plumbing functions and methods in e2e-framework as helpers too.

I didn't raise an issue tracking this.

```release-note
Internal machinery used to call step functions is now marked as test helpers. This means you can mark your step functions as a test helper using `t.Helper()`, and test logs will be include the filename and line number of `environment.Test()` call.
```

```docs

```

### DIFF
--- a/pkg/env/action.go
+++ b/pkg/env/action.go
@@ -72,6 +72,7 @@ type action struct {
 
 // runWithT will run the action and inject *testing.T into the callback function.
 func (a *action) runWithT(ctx context.Context, cfg *envconf.Config, t *testing.T) (context.Context, error) {
+	t.Helper()
 	switch a.role {
 	case roleBeforeTest, roleAfterTest:
 		if cfg.DryRunMode() {
@@ -98,6 +99,7 @@ func (a *action) runWithT(ctx context.Context, cfg *envconf.Config, t *testing.T
 
 // runWithFeature will run the action and inject a FeatureInfo object into the callback function.
 func (a *action) runWithFeature(ctx context.Context, cfg *envconf.Config, t *testing.T, fi types.Feature) (context.Context, error) {
+	t.Helper()
 	switch a.role {
 	case roleBeforeFeature, roleAfterFeature:
 		if cfg.DryRunMode() {
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -207,6 +207,7 @@ func (e *testEnv) panicOnMissingContext() {
 // processTestActions is used to run a series of test action that were configured as
 // BeforeEachTest or AfterEachTest
 func (e *testEnv) processTestActions(ctx context.Context, t *testing.T, actions []action) context.Context {
+	t.Helper()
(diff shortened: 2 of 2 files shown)


sttts:
Is this actually helping? `t.Helper()` just removes one frame of the stack. I.e. you see the file+line of the parent (or some ancestor) during execution of the closure returned by `funcs.SomeAssessFn()`, not during execution of `funcs.SomeAssessFn`. So you will not know which of the `funcs.SomeAssessFn` failed, but only the file (and test).

sttts:
I see that you comment exactly that in https://github.com/crossplane/crossplane/pull/5722#issuecomment-2148427387. So this PR is better than nothing clearly.

vladimirvivien (commented):
This is great @negz . Any side effects of calling t.Helper() successively on the same *testing.T value ?

cc @harshanarayana

vladimirvivien:
Apologies for the late review on this folks.

negz:
@vladimirvivien Not that I know of. I believe it's pretty normal to call it repeatedly (e.g. if you have several layers of test helper functions).

harshanarayana:
@vladimirvivien  Calling the `t.Helper` multiple times should be ok. It Should be using the current frames to do the analysis. 

```golang
package abc

import (
	"os"
	"testing"
)

func TestMain(m *testing.M) {
	os.Exit(m.Run())
}

func assertSomething(t *testing.T, a int) {
	t.Helper()
	t.Error("something failed")
}

func TestSomething(t *testing.T) {
	data := []int{
		1, 2, 3, 4, 5, 6, 7, 8, 9, 10,
	}

	for tc := range data {
		assertSomething(t, tc)
	}
}

harshanarayana (commented):
/lgtm

vladimirvivien:
This is failing consistently it seems, is there a regression ?

/retest

harshanarayana:
@vladimirvivien  seems like a flaky test. Let me take a look.

cpanato:
@negz can you please rebase and fix the conflicts? thanks

PR #424, opened 2024-06-05
execute test suite only if the setup was successful

### LINKED ISSUE #423: Test suite is executed even if the setup failed
### What happened?

The test suite is executed regardless of whether the setup actions were successful or not.

Tests should only be executed if all setup actions were successfully executed.

It can be reproduced by adding a failing setup action to the test environment, e.g.:
```
testEnv.Setup(
	func(ctx context.Context, _ *envconf.Config) (context.Context, error) {
		return ctx, errors.New("this is an error")
	},
)
```

_No response_

### E2E Provider Used

kind

### `e2e-framework` Version

v0.4.0

### OS version

_No response_

snorwin:
#### What type of PR is this?
/kind bug

Since the last changes ( #362) in the error handling of the setup functions, the test suite is also executed if the setup has failed. This can cause the tests to run to timeout because the expected prerequisites are not met. This PR introduces a check to ensure that the test suite is only executed if the setup was successful, otherwise the function returns with exit code 1 and the finish functions are gracefully executed.

Fixes #423

None

```docs
The test suite is only executed if the setup was successful.
```

```docs

```

### DIFF
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -382,18 +382,13 @@ func (e *testEnv) Run(m *testing.M) (exitCode int) {
 		// context passed down to each setup
 		if ctx, err = setup.run(ctx, e.cfg); err != nil {
 			klog.Errorf("%s failure: %s", setup.role, err)
-			exitCode = 1
-			break
+			return 1
 		}
 	}
 	e.ctx = ctx
 
-	// Execute the test suite (if the setup was successful)
-	if err == nil {
-		exitCode = m.Run()
-	}
-
-	return exitCode
+	// Execute the test suite
+	return m.Run()
 }
 
 func (e *testEnv) getActionsByRole(r actionRole) []action {
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -382,13 +382,18 @@ func (e *testEnv) Run(m *testing.M) (exitCode int) {
 		// context passed down to each setup
 		if ctx, err = setup.run(ctx, e.cfg); err != nil {
 			klog.Errorf("%s failure: %s", setup.role, err)
+			exitCode = 1
 			break
 		}
 	}
 	e.ctx = ctx
 
-	// Execute the test suite
-	return m.Run()
+	// Execute the test suite (if the setup was successful)
+	if err == nil {
+		exitCode = m.Run()
+	}
+
+	return exitCode
 }
 
 func (e *testEnv) getActionsByRole(r actionRole) []action {


snorwin:
Friendly ping for feedback to @harshanarayana  and @ShwethaKumbla .
Thanks in advance 🙏

snorwin:
@harshanarayana thanks for triggering the tests,  any chance that this PR is getting reviewed and released soon?

cpanato (commented):
/lgtm

vladimirvivien (commented):
This looks simple and it makes sense to be. But I think it will be impactful for existing users who may rely on that existing behavior. Are there cases where continuing with tests after a setup failure is good (I don't know, just asking). 

cc @harshanarayana @cpanato

snorwin:
@vladimirvivien The bug causing tests to run despite setup failures was introduced in v0.4.0 (commit: 70b0316). This caused test pipelines to timeout and generate strange error reports.

pkg/env/env.go
```diff
@@ -382,13 +382,18 @@ func (e *testEnv) Run(m *testing.M) (exitCode int) {
...
 	e.ctx = ctx
 
-	// Execute the test suite
-	return m.Run()
+	// Execute the test suite (if the setup was successful)
+	if err == nil {
+		exitCode = m.Run()
+	}
+
+	return exitCode
```
harshanarayana, line 396: Why not just `return 1` from line 385 instead ?

harshanarayana (commented):
@vladimirvivien  I think it should be ok. Like @snorwin  mentioned, this was a side effect of changes done in 70b0316dcdbe482ec16e81615df3491da0e0e11b. 

Which is really recent change. So, making this change will only revert it back to the original behavior of how things were running with a slight change that the cleanup will happen due to graceful termination unlike the Fatal one before.

pkg/env/env.go
snorwin: Thanks for your feedback. I thought it would be easier to read this way, since the context has to be assigned to the environment anyway: `e.ctx = ctx`.

I can change it to:
```
	if ctx, err = setup.run(ctx, e.cfg); err != nil {
			klog.Errorf("%s failure: %s", setup.role, err)
			e.ctx = ctx
			return 1
	}
	e.ctx = ctx
```

What do you think?

cpanato: amd why you assign the ctx to `e.ctx`

snorwin: ah this is actually not necessary, I thought `e.ctx` is used for the `Finish` actions, but it is the local copy, i.e., `ctx`.

snorwin:
@cpanato thanks for considering this fix. Is there a chance to get it released soon?

cpanato:
i will check what we have to be released, but no plans for this week

Issue #426, opened 2024-06-07
CLI flags for "running only given features" will skip whole test if a feature in the env doesn't match the criteria.


0sewa0:
### What happened?

When using the `--labels "key=value"` argument, the `Feature` with that label didn't run.
All `Features` got skipped, regardless of the label they have, because the first in the `Environment` didn't have the label.

### What did you expect to happen?

Run all the `Features` that have the label, and skip the ones that don't.

### How can we reproduce it (as minimally and precisely as possible)?

Have 2 `Features`, only second one has the label, try to run only the second one using the `--labels` arg.

### Anything elese we need to know?

Tried it with the `--feature` arg as well, same problem.

### E2E Provider Used

real cluster

### `e2e-framework` Version

v0.4.0

### OS version

<details>

```console
OS: MacOS 14.5 (Intel Machine)
```

</details>

0sewa0:
I am not familiar with the code(at all), and how it's structured, but out of curiosity, I went digging:

I think https://github.com/kubernetes-sigs/e2e-framework/blame/72aa5e3c117ef7c0a430328fa4b6e61edbdcf790/pkg/env/env.go#L209-L212 is incorrect.
- Doesn't `t.Skipf` just skip the whole test, not only the `Feature`? 🤔 
- Wouldn't a simple `return ctx` be a fix ? (with some log before it, of course 😅 )

PR #428, opened 2024-06-10
Skip only features that don't match the criteria

0sewa0:
#### What type of PR is this?
/kind bug
/kind regression

We only skip the `Feature` and not the whole test (ie.: All `Features` in a test `Env`, if 1 `Feature` fails to match the criteria).)

Just use `t.Log` + `return ctx` instead of `t.Skipf`:
- Problems with: `t.Skipf`
  - Skips the whole test => which will skip all other `Features` without checkin if they match the criteria.
  - Minor mistake: `t.Skipf` should be used if a template is used, but that is not the case here 🙃 

Fixes https://github.com/kubernetes-sigs/e2e-framework/issues/426

```release-note
```

```docs

```

### DIFF
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -208,7 +208,8 @@ func (e *testEnv) processTestActions(ctx context.Context, t *testing.T, actions 
 func (e *testEnv) processTestFeature(ctx context.Context, t *testing.T, featureName string, feature types.Feature) context.Context {
 	skipped, message := e.requireFeatureProcessing(feature)
 	if skipped {
-		t.Skipf(message)
+		t.Log(message)
+		return ctx
 	}
 	// execute beforeEachFeature actions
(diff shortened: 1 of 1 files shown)


pkg/env/env.go
```diff
@@ -208,7 +208,8 @@ func (e *testEnv) processTestActions(ctx context.Context, t *testing.T, actions
 func (e *testEnv) processTestFeature(ctx context.Context, t *testing.T, featureName string, feature types.Feature) context.Context {
 	skipped, message := e.requireFeatureProcessing(feature)
 	if skipped {
-		t.Skipf(message)
+		t.Log(message)
+		return ctx
```
harshanarayana, line 212: This will not mark the test as `SKIP`'ed correct ? The intent of `skip` here was to make it clear that the test was skipped. Returning ctx will not mark the test as skipped. Can be misleading.

PR #439, opened 2024-07-12
envconf: use crypto/rand for random name generator

harshanarayana:
#### What type of PR is this?

/kind bug

Currently, the method used to generate random name was relying on `math/rand`. Switching that to `crypto/rand` provide more randomness and better behavior. 

Though it is not that important to use cryptographically secure randomness, it is better to do that and avoid possible collisions. 

Fixes #

Added a test to account for and validate possible duplicate of random name generation. 

```release-note
switched to using `crypto/rand` instead of `math/rand` for generating random names using `envfunc.RandomName` handler. 
```

```docs

```


cpanato (approved):
thanks

/lgtm

PR #450, opened 2024-08-29
feat: add vcluster e2e cluster provider

crandles:
addition of vcluster provider; interested to see if the CI fails. vcluster needs a "host" cluster to launch into, so I've followed an existing example where a real cluster or kind cluster are utilized as necessary: https://github.com/kubernetes-sigs/e2e-framework/blob/9206e6f6082907a1c77750b7cac0fc44ff4dac02/examples/real_cluster/main_test.go#L39

#### What type of PR is this?

/kind feature

This PR enables a new `E2EClusterProvider` for `vcluster` based clusters.

Fixes #

```release-note
NONE
```

```docs
NA
```

### DIFF
--- a/third_party/vcluster/vcluster.go
+++ b/third_party/vcluster/vcluster.go
@@ -1,9 +1,12 @@
 /*
 Copyright 2024 The Kubernetes Authors.
+
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
+
     http://www.apache.org/licenses/LICENSE-2.0
+
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/third_party/vcluster/vcluster.go
+++ b/third_party/vcluster/vcluster.go
@@ -40,7 +40,7 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
-var vclusterVersion = "v0.20.0"
+var vclusterVersion = "v0.21.0"
 
 type Cluster struct {
 	path            string
@@ -273,7 +273,7 @@ func (c *Cluster) KubernetesRestConfig() *rest.Config {
 // helpers to implement support.E2EClusterProvider
 func (c *Cluster) findOrInstallVcluster() error {
 	version := c.version
-	if c.version != "" {
+	if c.version == "" {
 		version = vclusterVersion
 	}
 	path, err := utils.FindOrInstallGoBasedProvider(c.path, "vcluster", "github.com/loft-sh/vcluster/cmd/vcluster", version)
--- a/third_party/vcluster/vcluster.go
+++ b/third_party/vcluster/vcluster.go
@@ -40,7 +40,10 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
-var vclusterVersion = "v0.21.0"
+const (
+	vclusterVersion = "v0.20.0"
+	vclusterPath    = "vclusterctl"
+)
 
 type Cluster struct {
 	path            string
@@ -68,7 +71,7 @@ func WithPath(path string) support.ClusterOpts {
 	return func(c support.E2EClusterProvider) {
 		v, ok := c.(*Cluster)
 		if ok {
-			v.path = path
+			v.WithPath(path)
 		}
 	}
 }
@@ -124,7 +127,7 @@ func (c *Cluster) WithOpts(opts ...support.ClusterOpts) support.E2EClusterProvid
 
 func (c *Cluster) SetDefaults() support.E2EClusterProvider {
 	if c.path == "" {
-		c.path = "vcluster"
+		c.path = vclusterPath
 	}
 	return c
 }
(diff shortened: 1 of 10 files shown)


vladimirvivien:
@crandles this is awesome and thanks for the contribution.
At first glance it looks Okay, but I will ask @harshanarayana to take a look as well since he's the E2E provider expert 😉

third_party/vcluster/vcluster.go
```diff
@@ -0,0 +1,324 @@
...
+	if len(args) > 0 {
+		command = fmt.Sprintf("%s %s", command, strings.Join(args, " "))
+	}
+	log.V(4).Info("Launching:", command)
+	echo := gexe.New()
+	if c.hostKubeCfg != "" {
+		echo.SetEnv("KUBECONFIG", c.hostKubeCfg)
+	}
+
+	p := echo.RunProc(command)
```
vladimirvivien, line 159: I think @harshanarayana has some util wrapper for running procs that should be considered here.

third_party/vcluster/vcluster.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+	return c
+}
+func (c *Cluster) SetDefaults() support.E2EClusterProvider {
+	if c.path == "" {
+		c.path = "vcluster"
+	}
+	return c
+}
+func (c *Cluster) Create(ctx context.Context, args ...string) (string, error) {
```
vladimirvivien, line 126: @crandles does this vcluster create function will only work for kind "real" clusters (i don't know much about vclusters). If that is the case, you may wanna consider making that clear in naming (i.e. third_party/kind_vcluster).

vladimirvivien (commented):
This is a great start. I left some comments.

third_party/vcluster/vcluster.go
crandles: No, this doesn't require kind, this should work in any host k8s cluster.

I set it up via kind in the example so that it could presumably execute successfully (doubting there's a real cluster to access otherwise within CI).

crandles:
golangci appears to be passing now for thew new code I've added, though errirng for some existing LOC:

```
third_party/helm/helm.go:240:9: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
		err = fmt.Errorf(missingHelm)
		      ^
third_party/flux/flux_setup.go:47:16: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
			return ctx, fmt.Errorf(NoFluxInstallationFoundMsg)
			            ^
third_party/flux/flux_setup.go:61:16: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
			return ctx, fmt.Errorf(NoFluxInstallationFoundMsg)
			            ^
pkg/env/env.go:229:11: printf: non-constant format string in call to (*testing.common).Skipf (govet)
		t.Skipf(message)
		        ^
pkg/env/env.go:513:22: printf: non-constant format string in call to (*testing.common).Skipf (govet)
					internalT.Skipf(message)
```

Assuming that won't block this PR.

vladimirvivien:
@harshanarayana and @cpanato any idea why golangci failing on previously working code ?

crandles:
errors were primarily for:
> SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)

It's possible that I'm using a different toolchain version, but as the warning seemed fair, I've fixed it via https://github.com/kubernetes-sigs/e2e-framework/pull/450/commits/cf9cbf28dfc7a407ed901b2da26b31ef4aabcc35

vladimirvivien:
@crandles is this ready to go now ?

crandles:
Yes, I believe so

crandles:
rebased

vladimirvivien (approved):
LGTM

vladimirvivien:
@harshanarayana @ShwethaKumbla @cpanato PTAL I think we need a second reviewer.

harshanarayana:
I will check by eod tomorrow.

crandles:
Seems to fail installing vcluster; can't reproduce. Any tips? I think I'm installing vcluster the same way other tools are (ko|kwok|kind)

harshanarayana:
> Seems to fail installing vcluster; can't reproduce. Any tips? I think I'm installing vcluster the same way other tools are (ko|kwok|kind)

Helm and flux are installed slightly in a different way, https://github.com/kubernetes-sigs/e2e-framework/blob/777cbcd59fbbe572b71dde89c8f7b0b1a5ddc4a7/hack/install-helm.sh

```bash
test: install-helm install-flux ## Runs golang unit tests
	./hack/test-go.sh

crandles:
Is there a recommended way to simulate the prow test runner locally?

This looks relevant: https://docs.prow.k8s.io/docs/build-test-update/#running-a-prowjob-locally, I'll try to follow it.

crandles:
found a couple of issues:
- I was mistaken on the vcluster installation path, it is vclusterctl not vcluster
  - Apparently I've been falling back to an alternatively installed vcluster in my CI
- there was a flaw in the version defaulting logic

cpanato (approved):
thanks

/lgtm
/approve

crandles:
Thanks for the reviews!

I will try to rebase next time, and I may follow up with an issue about how to run the pr tests locally. I need to spend more time looking at the pj on kind script.

PR #453, opened 2024-09-16
feat: add benchmark and skipping features example

### LINKED ISSUE #268: Allow execution of non-feature tests
Support the execution of tests without requiring use of the [features](https://pkg.go.dev/sigs.k8s.io/e2e-framework@v0.2.0/pkg/features) package.

To the [Environment](https://github.com/kubernetes-sigs/e2e-framework/blob/v0.2.0/pkg/internal/types/types.go#L49-L88) interface, I propose the addition of two new functions:

```go
type Environment interface {
	RunTest(ctx context.Context, t *testing.T, cfg *envconf.Config) bool
	RunBenchmark(ctx context.Context, b *testing.B, cfg *envconf.Config) bool
}
```

I am unsure what if anything makes sense as return values, [t.Run](https://pkg.go.dev/testing?utm_source=godoc#T.Run) returns a bool which may fit here; additionally `RunBenchmark` might deserve its own issue, given the current lack of support overall for benchmarks.

Note, there is some overlap in wording with this existing function, unsure if the above naming will be confusing:
```go
	// Run Launches the test suite from within a TestMain
	Run(*testing.M) int
```

The goal is to allow a more stream-lined path to writing go tests natively or otherwise with the framework of choice; don't force usage of the features package.

Potential usage:
```go
testenv.RunTest(ctx context.Context, t *testing.T, cfg *envconf.Config) (context.Context, error){
  ...
  if err != nil {
    t.Fatal(err)
  }
  ...
   t.Run("...", func(t testing.T){
     t.Parallel()
     ...
   })

})
```

crandles:
#### What type of PR is this?

/kind documentation

This adds an example of skipping the `features` package and how to run a benchmark using the `testenv` client.

Fixes #268

```release-note
NONE
```

See included README


cpanato (commented):
/ok-to-test

vladimirvivien (commented):
This is great @crandles . 
@cpanato @harshanarayana PTAL

vladimirvivien (approved):
LGTM

vladimirvivien:
@cpanato @harshanarayana PTAL

PR #471, opened 2025-01-03
Extend GetKubeconfig methods with args

mjudeikis:
#### What type of PR is this?

/kind feature

When dealing with multi-cluster setup we ran into issue where we need to pass arguments into `GetKubeconfig` method. 
Specific example in our case was `--internal` to `kind` so cluster can speak to different clusters.

This extends this so you can do things like:
```
internalKubeConfigOpts := kind.WithGetKubeConfigArgs("--internal")
		internalKubeConfig, err := kindEdgeCluster.WithOpts(internalKubeConfigOpts).GetLiveKubeconfig()
		if err != nil {
			log.Printf("Failed to get edge cluster kubeconfig: %s", err)
			return ctx, err
		}
		installCmd := fmt.Sprintf(
			`bash -c "KUBECONFIG=%s kubectl create secret generic edge-cluster-kubeconfig -n omni-operator-system  --from-file=kubeconfig=%s"`,
			kindMasterCluster.GetKubeconfig(),
			internalKubeConfig,
		)
```

Or similar pattern if you want to set this for all `cluster` object. 
In addition to allow networking we did this:
```
os.Setenv("KIND_EXPERIMENTAL_DOCKER_NETWORK", "kind-test")
	testEnv.Setup(
		envfuncs.CreateCluster(kindMasterCluster, kindMasterClusterName),
		envfuncs.CreateNamespace(namespace),
		setupFunc,
		installDepsFunc,
		generateAndDeployFunc,
		envfuncs.CreateClusterWithConfig(kindEdgeCluster, kindEdgeClusterName, "edge-kind.yaml"), // must be after master cluster setup.
		joinMasterClusterPrerequisitesFunc,
	)
```

Im not super exited about naming of `GetLiveKubeconfig` so any suggestions are welcome 

Fixes #

```release-note
```

```docs

```

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -196,6 +196,11 @@ type E2EClusterProvider interface {
 	// using the cluster provider native way
 	GetKubeconfig() string
 
+	// GetLatestKubeconfig provides a way to extract the kubeconfig file associated with the cluster in question.
+	// Different from GetKubeconfig, this method is used to extract the kubeconfig using provider native way, where the GetKubeconfig is cached.
+	// It is useful in cases where the kubeconfig file is required to be extracted multiple times during the test run with different arguments.
+	GetLatestKubeconfig(args ...string) (string, error)
+
 	// GetKubectlContext is used to extract the kubectl context to be used while performing the operation
 	GetKubectlContext() string
 
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -196,10 +196,10 @@ type E2EClusterProvider interface {
 	// using the cluster provider native way
 	GetKubeconfig() string
 
-	// GetLatestKubeconfig provides a way to extract the kubeconfig file associated with the cluster in question.
+	// GenerateKubeconfig provides a way to extract the kubeconfig file associated with the cluster in question.
 	// Different from GetKubeconfig, this method is used to extract the kubeconfig using provider native way, where the GetKubeconfig is cached.
 	// It is useful in cases where the kubeconfig file is required to be extracted multiple times during the test run with different arguments.
-	GetLatestKubeconfig(args ...string) (string, error)
+	GenerateKubeconfig(args ...string) (string, error)
 
 	// GetKubectlContext is used to extract the kubectl context to be used while performing the operation
 	GetKubectlContext() string
--- a/third_party/k3d/k3d.go
+++ b/third_party/k3d/k3d.go
@@ -251,7 +251,7 @@ func (c *Cluster) GetKubectlContext() string {
 	return fmt.Sprintf("k3d-%s", c.name)
 }
 
-func (c *Cluster) GetLatestKubeconfig(args ...string) (string, error) {
+func (c *Cluster) GenerateKubeconfig(args ...string) (string, error) {
 	return c.getKubeConfig(args...)
 }
 
--- a/third_party/k3d/k3d.go
+++ b/third_party/k3d/k3d.go
@@ -106,11 +106,12 @@ func (c *Cluster) findOrInstallK3D() error {
 	return err
 }
 
-func (c *Cluster) getKubeConfig() (string, error) {
+func (c *Cluster) getKubeConfig(args ...string) (string, error) {
 	kubeCfg := fmt.Sprintf("%s-kubecfg", c.name)
 
 	var stdout, stderr bytes.Buffer
-	err := utils.RunCommandWithSeperatedOutput(fmt.Sprintf("%s kubeconfig get %s", c.path, c.name), &stdout, &stderr)
+	cmd := fmt.Sprintf("%s kubeconfig get %s %s", c.path, strings.Join(args, " "), c.name)
+	err := utils.RunCommandWithSeperatedOutput(cmd, &stdout, &stderr)
 	if err != nil {
 		return "", fmt.Errorf("failed to get kubeconfig: %s", stderr.String())
 	}
@@ -248,6 +249,10 @@ func (c *Cluster) GetKubeconfig() string {
 
 func (c *Cluster) GetKubectlContext() string {
 	return fmt.Sprintf("k3d-%s", c.name)
+}
+
+func (c *Cluster) GetLatestKubeconfig(args ...string) (string, error) {
+	return c.getKubeConfig(args...)
 }
 
 func (c *Cluster) ExportLogs(ctx context.Context, dest string) error {
--- a/third_party/kind/kind.go
+++ b/third_party/kind/kind.go
@@ -308,6 +308,6 @@ func (k *Cluster) KubernetesRestConfig() *rest.Config {
 	return k.rc
 }
 
-func (k *Cluster) GetLatestKubeconfig(args ...string) (string, error) {
+func (k *Cluster) GenerateKubeconfig(args ...string) (string, error) {
 	return k.getKubeconfig(args...)
 }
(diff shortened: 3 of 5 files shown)


support/kwok/kwok.go
```diff
@@ -89,8 +100,11 @@ func (k *Cluster) getKubeconfig() (string, error) {
 	kubecfg := fmt.Sprintf("%s-kubecfg", k.name)
 
 	var stdout, stderr bytes.Buffer
-	err := utils.RunCommandWithSeperatedOutput(fmt.Sprintf(`%s get kubeconfig --name %s`,
-		k.path, k.name), &stdout, &stderr)
+	cmd := fmt.Sprintf(`%s get kubeconfig --name %s`, k.path, k.name)
+	if len(k.getKubeconfigArgs) > 0 {
+		cmd = fmt.Sprintf("%s %s", cmd, strings.Join(k.getKubeconfigArgs, " "))
+	}
```
dmvolod, line 106: strings.Join() already has validation your provided and seems to it's not needed here
```
// Join concatenates the elements of its first argument to create a single string. The separator
// string sep is placed between elements in the resulting string.
func Join(elems []string, sep string) string {
	switch len(elems) {
	case 0:
		return ""
```

```suggestion
	cmd := fmt.Sprintf(`%s get kubeconfig %s --name %s`, k.path, strings.Join(k.getKubeconfigArgs, " "), k.name)
```

mjudeikis: Thanks. Lets tests githubs and prow's ability to deal with your propose change :)

dmvolod: Oh, current tests are flaky, I've never seen to pass all tests.

mjudeikis:
And I ran into this again. Karma...  :D rebased and repushed

pkg/types/types.go
```diff
@@ -196,6 +202,11 @@ type E2EClusterProvider interface {
 	// using the cluster provider native way
 	GetKubeconfig() string
 
+	// GetLiveKubeconfig provides a way to extract the kubeconfig file associated with the cluster in question.
+	// Different from GetKubeconfig, this method is used to extract the kubeconfig live, where the GetKubeconfig is cached.
+	// It is useful in cases where the kubeconfig file is required to be extracted multiple times during the test run with different arguments.
+	GetLiveKubeconfig() (string, error)
```
vladimirvivien, line 208: Not clear what the difference is. between `GetKubeConfig` and `GetLiveKubeConfig`  . It looks like in each implementation, it's just a call to `c.getKubeConfig`.

vladimirvivien (commented):
Hi this is a good start. I left some comments. Also I would ask to update any example or documentation to indicate this change.

pkg/types/types.go
```diff
@@ -179,6 +179,12 @@ type E2EClusterProvider interface {
 	// PATH variable and you want to use that instead of framework trying to install one on it's own.
 	WithPath(path string) E2EClusterProvider
 
+	// WithGetKubeConfigArgs is used to provide a mechanism where the cluster provider can be configured
+	// to provide additional flags to the get kubeconfig command that is used to extract the kubeconfig
+	// file from the cluster. This can be used to provide additional flags such as --internal for kind.
+	// It is not validated by the framework and is passed as is to the cluster provider.
+	WithGetKubeConfigArgs(args ...string) E2EClusterProvider
```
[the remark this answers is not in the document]
mjudeikis: This was argument before to add new one over changing existing one. Happy to change it if we ok with this change

pkg/types/types.go
mjudeikis: I think I can remove this. I needed this at the start of the year (wrapper around same interface outside this code), but now use-case is slightly different and I will remove it.

mjudeikis: Ok, I see the problem now (still there)
```
func (c *Cluster) GetKubeconfig() string {
	return c.kubeConfigFile
}

func (c *Cluster) GetLiveKubeconfig() (string, error) {
	return c.getKubeConfig()
}
```
Basically `GetKubeconfig` always returns kubeconfig file, created during setup in tmp dir, where live will re-run get command and potentially re-apply args (might be changed with With helper)

So either we need to stop returning cached kubeconfig and always get fresh or still have this method :/

pkg/types/types.go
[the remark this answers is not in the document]
vladimirvivien: See comment below.

vladimirvivien (commented):
Added a new comment, getting close.

pkg/types/types.go
mjudeikis: `Latest` it is :) will update

pkg/types/types.go
vladimirvivien, line 186: One more thing about the constructor function: instead of `WithGetKubeConfigArgs` i would suggest `WithKubeConfigArgs` or simply `WithArgs` (recommended, which would be a way to pass arbitrary args to the E2EProvider)

vladimirvivien (commented):
One more suggested change and I think it will be ready. Thanks for your patience.

pkg/types/types.go
vladimirvivien: @mjudeikis I understand what you are trying to do with that constructor function. However, either it's not appropriate to do it as a constructor function and/or the naming is not clear. Furthermore, with this PR, there will be `2` methods `GetKubeConfig()` and `GetLatestKubeConfig()`.  It's not clear which of the two will use these arguments.

### Suggestion 1: Move arguments to the method
I would suggest to move the arguments at the places where they are needed:

```go
GetLatestKubeconfig(args ...string) (string, error)
```
#### Suggestion 2: General `ClusterProvider` Args
I think having a constructor function named `WithArgs` is appropriate as it makes it possible to inject `raw` parameters into the construction of the ClusterProvider. One way this could be done is to use an Optional design:

```go
type argOptions struct{
    kubeCfgArgs []string
}

type ArgOption func(*argOptions)

func WithArgs(opt ... ArgOption)
```
So that you can have different types of arguments that is passed.  So for the kubeconfig retrieval, it could look something like:

```go
func WithKubeconfigArgs(args ...string) ArgOption {
    return func(opts *argOptions) {
        opts.kubCfgArgs = args...
    }
}

provider := WithArgs(WithKubeconfigArgs("arg1", "val1", "arg2", "val2"))
```
This now gives E2EClusterProvider implementer the flexibility to add new types of arguments in the future.

vladimirvivien (commented):
Responded to last comment.

pkg/types/types.go
mjudeikis: I think 1st is more reasonable here. Because we have ClusterOpts already and wit would be very much confusing what is the difference. In reality there is none. One could implement this already (Add opts I mean, not getKubeConfig)

pkg/types/types.go
```diff
@@ -196,6 +196,11 @@ type E2EClusterProvider interface {
 	// using the cluster provider native way
 	GetKubeconfig() string
 
+	// GetLatestKubeconfig provides a way to extract the kubeconfig file associated with the cluster in question.
+	// Different from GetKubeconfig, this method is used to extract the kubeconfig using provider native way, where the GetKubeconfig is cached.
+	// It is useful in cases where the kubeconfig file is required to be extracted multiple times during the test run with different arguments.
+	GetLatestKubeconfig(args ...string) (string, error)
```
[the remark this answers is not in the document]
vladimirvivien: @mjudeikis let me know what you think so we can get this merged.

Issue #472, opened 2025-01-03
Skipping when fail-fast is set results in failure


phisco:
### What happened?

See reproducer [here](https://github.com/phisco/e2e-framework-test-skip).

Running the test defined results in the whole suite failing in the latest releases, both [0.4.0](https://github.com/phisco/e2e-framework-test-skip/pull/1) and 0.5.0 on main branch:
```
go test -v -test.count 1 ./...
=== RUN   TestSkip
=== RUN   TestSkip/skip
=== RUN   TestSkip/skip/Assess_1
    main_test.go:23: Assess 1 (should be printed)
    main_test.go:24: skipping Assess 1
--- FAIL: TestSkip (0.00s)
    --- FAIL: TestSkip/skip (0.00s)
        --- SKIP: TestSkip/skip/Assess_1 (0.00s)
FAIL
FAIL    github.com/phisco/e2e-framework-test-skip       0.497s
FAIL
```

### What did you expect to happen?

Just the assessment `Assess_1` to be skipped, but the suite to succeed as in `0.3.0`, see [PR](https://github.com/phisco/e2e-framework-test-skip/pull/2) .

```
go test -v -test.count 1 ./...
=== RUN   TestSkip
=== RUN   TestSkip/skip
=== RUN   TestSkip/skip/Assess_1
    main_test.go:23: Assess 1 (should be printed)
    main_test.go:24: skipping Assess 1
=== RUN   TestSkip/skip/Assess_2
    main_test.go:28: Assess 2 (should be printed)
=== RUN   TestSkip/succeed
=== RUN   TestSkip/succeed/Assess_1
    main_test.go:35: Assess 1 (should be printed)
=== RUN   TestSkip/succeed/Assess_2
    main_test.go:39: Assess 2 (should be printed)
--- PASS: TestSkip (0.00s)
    --- PASS: TestSkip/skip (0.00s)
        --- SKIP: TestSkip/skip/Assess_1 (0.00s)
        --- PASS: TestSkip/skip/Assess_2 (0.00s)
    --- PASS: TestSkip/succeed (0.00s)
        --- PASS: TestSkip/succeed/Assess_1 (0.00s)
        --- PASS: TestSkip/succeed/Assess_2 (0.00s)
PASS
ok      github.com/phisco/e2e-framework-test-skip       0.467s
```

### How can we reproduce it (as minimally and precisely as possible)?

See reproducer [here](https://github.com/phisco/e2e-framework-test-skip).

[...]

PR #473, opened 2025-01-03
fix: properly handle skip on assessments

phisco:
#### What type of PR is this?

/kind bug

`t.Skip` with `--fail-fast` was resulting in the parent test being failed, this patch fixes that, however I'm not entirely sure about the sematic of `--fail-fast`, so I'd appreciate if someone could clarify that.

For sure, independently of `--fail-fast` a call to `t.Skip*` in an `Assessment` should only skip the current Assessment. But in case of calls to `t.Fail` or `t.FailNow`, what should be the behaviour in the two cases?

Fixes https://github.com/kubernetes-sigs/e2e-framework/issues/472

```release-note
Fix bug failing tests being skipped when `--fail-fast` is set.
```

```docs

```

### DIFF
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -27,7 +27,7 @@ import (
 	"sync"
 	"testing"
 
-	klog "k8s.io/klog/v2"
+	"k8s.io/klog/v2"
 
 	"sigs.k8s.io/e2e-framework/klient"
 	"sigs.k8s.io/e2e-framework/pkg/envconf"
@@ -480,6 +480,7 @@ func (e *testEnv) executeSteps(ctx context.Context, t *testing.T, steps []types.
 func (e *testEnv) execFeature(ctx context.Context, t *testing.T, featName string, f types.Feature) context.Context {
 	t.Helper()
 	// feature-level subtest
+	var failFast bool
 	t.Run(featName, func(newT *testing.T) {
 		newT.Helper()
 
@@ -494,7 +495,6 @@ func (e *testEnv) execFeature(ctx context.Context, t *testing.T, featName string
 		// assessments run as feature/assessment sub level
 		assessments := features.GetStepsByLevel(f.Steps(), types.LevelAssess)
 
-		failed := false
 		for i, assess := range assessments {
 			assessName := assess.Name()
 			if dAssess, ok := assess.(types.DescribableStep); ok && dAssess.Description() != "" {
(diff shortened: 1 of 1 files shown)


Fricounet (commented):
Thanks for catching and fixing the bug.

> however I'm not entirely sure about the sematic of --fail-fast, so I'd appreciate if someone could clarify that.

I don't have the full context but from my understanding, `--fail-fast` is an attempt to adapt the [go test](https://go.dev/doc/go1.10#test) `-failfast` flag to the e2e-framework, ie have a way to stop assessments and features executions early.

> For sure, independently of --fail-fast a call to t.Skip* in an Assessment should only skip the current Assessment. But in case of calls to t.Fail or t.FailNow, what should be the behaviour in the two cases?

In my understanding of the framework, Assessments can be kinda linked together as part of the same 'Test' while features are some kind of distinct 'Test'. 
Based on the [semantics](https://pkg.go.dev/testing#T.Fail) of `Fail` (`marks the function as having failed but continues execution.`) and `FailNow` (`marks the function as having failed and stops its execution by calling runtime.Goexit. Execution will continue at the next test or benchmark.`), my interpretation is that:
- a call to `Fail` should not attempt to skip any remaining assessment or feature
- a call to `FailNow` should skip all the remaining assessment but not skip the next features

I guess the behavior for `FailNow` could be argued for or against skipping the next features too. But this is more a question of whether features should be considered independent or not, which, as far as I know is not really clarified anywhere.

phisco:
> - a call to Fail should not attempt to skip any remaining assessment or feature
> - a call to FailNow should skip all the remaining assessment but not skip the next features

Makes sense, @Fricounet, but what about `--fail-fast`? Should we skip the remaining assessments independently of which one was called, `Fail` or `FailNow` when `--fail-fast` is set? I'd say so from: https://github.com/kubernetes-sigs/e2e-framework/blob/b69e15831920a55e0c58ece5bf765594be7ecfdc/examples/fail_fast/README.md#L3-L11

Fricounet:
> Makes sense, @Fricounet, but what about --fail-fast? Should we skip the remaining assessments independently of which one was called, Fail or FailNow when --fail-fast is set? I'd say so from:

I'd say so too @phisco. Though, I don't agree with everything written in the README. In particular:
> This can aid in getting a better turn around time especially in cases where your assessments are 
 inter-related and a failed assessment in step 1 can mean that running the rest of the assessment 
 is guaranteed to fail. This can be done with the help of a `--fail-fast` flag provided at the 
 framework level. 

To me, this is the job of `t.FailNow()`

Basically:
- as the test writer, I know when it makes sense for my test to keep going (use `t.Fail()`) or stop early because there is no point going further (use `t.FailNow()`)
- as the test executor, I know if I want my whole test suite to exit early at the first sign of failure (use `--fail-fast`). Hence `--fail-fast` should skip the remaining assessments regardless of Fail/FailNow

Hope that makes sense :smile:

vladimirvivien:
@phisco @Fricounet @harshanarayana 

This PR slipped, but I think it's an important one to get clarity on and merge eventually.

Features such as `--fail-fast` as a way to unblock the framework's users, not knowing the full implications.
Now that the framework as been enjoying wider adoption, I think we can revisit that feature to ensure it's working the best way possible.

So, here is what i am proposing (which aligns with what @Fricounet mentioned):

- A test writer should use exit methods (`t.Fail`/`t.FailNow`/`t.Fatal`) for granular control of test steps 
- Call to exit methods should be applied to associated assessment/feature only, allowing remaining tests to continue
- `--fail-fast` - short-circuits current and all remaining tests (features, and assessments) if an exit method is called 
- `--fail-fast` also short-circuits the lifecycle of the entire framework runtime if a structural error occurs (i.e. during initialization, setup step)

Hope this clears thing up. If not, let me know what I am missing or what you want to add.

vladimirvivien:
@phisco are you still interested in continuing with this change?

PR #475, opened 2025-01-05
Add specific DecodeURL function in decoder package

### LINKED ISSUE #474: Add specific DecodeURL function in decoder package
### What do you want to see?

As developer, I would like to have out-of the-box solution to stream documents from the URL and decode them.
Yes, most of the functions utilize `io.Reader` and can be wrapped, but separate specific function is much better to use.
This will help in situations where we need to install components from URL, such as Prometheus or CertManager in examples.

### Extra Labels

_No response_

dmvolod:

Fixes #474 

```release-note
"NONE"
```

```docs
DecodeURL is a stream of documents from the URL of any Kind using either the innate typing of the scheme.
Falls back to the unstructured.Unstructured type if a matching type cannot be found for the Kind.
```

### DIFF
--- a/klient/decoder/decoder.go
+++ b/klient/decoder/decoder.go
@@ -27,6 +27,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/vladimirvivien/gexe/http"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


klient/decoder/decoder.go
```diff
@@ -217,6 +218,18 @@ func DecodeFile(fsys fs.FS, manifestPath string, obj k8s.Object, options ...Deco
 	return Decode(f, obj, options...)
 }
 
+// DecodeURL is a stream of documents from the URL of any Kind using either the innate typing of the scheme.
```
vladimirvivien, line 221: I would reword the start of the comment `// DecodeURL decodes a document from the URL...`

vladimirvivien (commented):
Thanks @dmvolod. Added a suggestion.

klient/decoder/decoder.go
dmvolod: Thanks, fixed.

vladimirvivien (approved):
/LGTM

PR #485, opened 2025-01-17
provider: enable k3d cluster log export helper

harshanarayana:
#### What type of PR is this?
/kind feature

Enable `k3d` provider with Ability to export cluster logs.  Changes required to enable this was added to `k3d` with https://github.com/k3d-io/k3d/pull/1471. 

With that changes being merged, we can now add support for exporting the logs. 

NA

```release-note
added ability to export cluster logs for `k3d` based Cluster Provider
```

```docs

```

### DIFF
--- a/third_party/k3d/k3d.go
+++ b/third_party/k3d/k3d.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/blang/semver/v4"
 	"k8s.io/apimachinery/pkg/util/json"
 
 	"k8s.io/client-go/rest"
@@ -37,7 +38,8 @@ import (
 	log "k8s.io/klog/v2"
 )
 
-var k3dVersion = "v5.7.2"
+var k3dVersion = "v5.8.0"
+var k3dWithExportLogSupport = "v5.8.0"
 
 type Cluster struct {
 	path           string
@@ -88,7 +90,7 @@ func WithImage(image string) support.ClusterOpts {
 }
 
 func NewCluster(name string) *Cluster {
-	return &Cluster{name: name}
+	return &Cluster{name: name, version: k3dVersion}
 }
 
 func NewProvider() support.E2EClusterProvider {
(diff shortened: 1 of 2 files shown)


harshanarayana:
/assign @cpanato 

PTAL when you can. @vladimirvivien @cpanato

harshanarayana (commented):
/hold until I finish validating this and add an Example for using this

cpanato (commented):
thanks

/lgtm

third_party/k3d/k3d.go
```diff
@@ -251,8 +253,25 @@ func (c *Cluster) GetKubectlContext() string {
...
+	if reqVersion.GE(supVersion) {
+		p := utils.RunCommandWithCustomWriter(fmt.Sprintf("%s debug export-logs %s --path %s", c.path, c.name, dest), &stdout, &stderr)
+		err = p.Err()
+		if err != nil {
+			log.ErrorS(err, "failed to export cluster logs due to an error", "stdout", stdout.String(), "stderr", stderr.String(), "result", p.Result())
+			return err
+		}
+		return nil
+	} else {
+		log.Warning("ExportLogs not implemented for k3d. Please use regular kubectl like commands to extract the logs from the cluster")
```
vladimirvivien, line 272: Do we know what that `kubectl` command would be to extract logs? Could we just use it as a fallback instead of just reporting a warning?

harshanarayana: @vladimirvivien  Actually, That `kubectl like` is probably a mistake. Just `kubectl` won't be enough. We will also have to get a few more logs. And the problem is, we have to find a way to group them in a proper format. it can be tricky. (Also a duplicate work to replicate the logic).  

If we want to add an export mechanism, we might be better off writing some mechanism that can do this for any provider that doesn't have a log export support. 

`vcluster` doesn't support. So is the case with `kwok`. 

For `kwok` there is an adhoc `kubectl logs` equivalent of export but it does only for the `kwok`'s own components. 

How do we go about doing this ?

harshanarayana: We also have to worry about fetching the journalctl logs + the dmesg + any other applicable logs. Might not be a bad idea to implement something generic that can be reused for any provider in the future that is lacking the log export capabilities

harshanarayana: @vladimirviven ptal.. I would love tp get this one merged.

PR #506, opened 2025-05-31
Add ExecInDeployment method

keeprocking:
#### What type of PR is this?

/kind feature

This patch adds the `ExecInDeployment` method, which works similarly to
```shell
$ kubectl exec -n <namespace> -it deploy/<deployment> -- <command>
```

It's useful when you need to run a command from one of the pods belonging to a specific deployment, but you don't care which one.

For example, suppose there's a deployment running a custom controller, and you need to assert metrics written by that controller. The metrics endpoint is exposed via a `Service`, so it doesn't really matter which pod you call it from.

Instead of manually retrieving the list of pods belonging to the deployment, selecting one and passing it to `ExecInPod`, you could do this:
```go
var stdout, stderr *bytes.Buffer
cmd := []string{"wget", "-qO-", fmt.Sprintf("http://my-service.%s.svc:8080/metrics", c.Namespace())}
if err := c.Client().Resources().ExecInDeployment(ctx, c.Namespace(), deployment.Name, cmd, &stdout, &stderr); err != nil {
  t.Log(stderr.String())
  t.Fatal(err)
}
metrics := stdout.String()
```

```release-note
Added the Resources.ExecInDeployment method, which finds a deployment by name and runs the command in the first container of the first pod in that deployment
```

```docs
```

### DIFF
--- a/klient/k8s/resources/resources.go
+++ b/klient/k8s/resources/resources.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"time"
 
+	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
@@ -327,6 +328,39 @@ func (r *Resources) ExecInPod(ctx context.Context, namespaceName, podName, conta
 	}
 
 	return nil
+}
+
+// ExecInDeployment runs the command in the first container of the first pod of the specified deployment,
+// mimicking the behavior of
+//
+// $ kubectl exec -it -n <namespaceName> deploy/<deploymentName> -- <command>
+func (r *Resources) ExecInDeployment(ctx context.Context, namespaceName, deploymentName string, command []string, stdout, stderr *bytes.Buffer) error {
+	var deployment appsv1.Deployment
+	if err := r.Get(ctx, deploymentName, namespaceName, &deployment); err != nil {
+		return err
+	}
+
+	sel, err := metav1.LabelSelectorAsSelector(deployment.Spec.Selector)
+	if err != nil {
+		return err
+	}
+
+	var pods v1.PodList
+	if err := r.client.List(ctx, &pods, cr.InNamespace(namespaceName), cr.MatchingLabelsSelector{Selector: sel}); err != nil {
+		return err
+	}
+
+	if len(pods.Items) == 0 {
+		return errors.New("deployment has no pods")
+	}
+	pod := pods.Items[0]
+
+	if len(pod.Spec.Containers) == 0 {
+		return errors.New("pod has no containers")
+	}
+	container := pod.Spec.Containers[0]
+
+	return r.ExecInPod(ctx, namespaceName, pod.Name, container.Name, command, stdout, stderr)
 }
 
 func init() {
--- a/examples/exec_in_deployment/README.md
+++ b/examples/exec_in_deployment/README.md
@@ -1,0 +1,23 @@
+# Execute Commands in a Deployment
+
+This package demonstrates how to use `ExecInDeployment` to execute commands inside a specific deployment.
+
+`ExecInDeployment` mimics the behavior of `kubectl exec -it -n <namespace> deploy/<deployment> -- <command>`: it selects the first pod of the specified deployment and runs the command in the first container of that pod.
+
+It provides a convenient way to run a command using only the deployment name — similar to `ExecInPod`, but for deployments, and defaulting to the first container of the first pod.
+
+`ExecInDeployment` expects the deployment with the specified name to exist and be available.
+
+## Example
+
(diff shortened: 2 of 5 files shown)


vladimirvivien:
Thanks for pinging me directly. I will take a look soon.

vladimirvivien (commented):
This is a great suggestion, but I left comments to make the usecase it a bit more general

klient/k8s/resources/resources.go
```diff
@@ -329,6 +330,39 @@ func (r *Resources) ExecInPod(ctx context.Context, namespaceName, podName, conta
 	return nil
 }
 
+// ExecInDeployment runs the command in the first container of the first pod of the specified deployment,
+// mimicking the behavior of
+//
+// $ kubectl exec -it -n <namespaceName> deploy/<deploymentName> -- <command>
+func (r *Resources) ExecInDeployment(ctx context.Context, namespaceName, deploymentName string, command []string, stdout, stderr *bytes.Buffer) error {
```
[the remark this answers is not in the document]
keeprocking: Should I keep the existing `ExecInDeployment`  or remove it at all?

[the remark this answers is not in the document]
vladimirvivien: @keeprocking I would recommend use either the two I suggested since they are more specific, to avoid confusion.

klient/k8s/resources/resources.go
```diff
@@ -329,6 +331,56 @@ func (r *Resources) ExecInPod(ctx context.Context, namespaceName, podName, conta
...
 
+type execInDeploymentOptions struct {
+	PodIndex int
+}
+
+// ExecInDeploymentOption extends the [ExecInDeployment] default behavior.
+type ExecInDeploymentOption func(*execInDeploymentOptions)
+
+// WithPodIndex sets the pod index to select.
+func WithPodIndex(idx int) ExecInDeploymentOption {
```
vladimirvivien, line 342: Because this package has many concerns, I would recommend making options names specific i.e. `WithPodIdxInDeployment` . Or, a more complex change, to create a new package under `./klient/k8s` for better namespacing.

klient/k8s/resources/resources.go
```diff
@@ -329,6 +331,56 @@ func (r *Resources) ExecInPod(ctx context.Context, namespaceName, podName, conta
 	return nil
 }
 
+type execInDeploymentOptions struct {
+	PodIndex int
```
vladimirvivien, line 335: This var should be unexported.

klient/k8s/resources/resources.go
```diff
@@ -329,6 +331,56 @@ func (r *Resources) ExecInPod(ctx context.Context, namespaceName, podName, conta
...
+
+	if options.PodIndex >= len(pods.Items) {
+		return fmt.Errorf("pod index %d is out of range (available pods: %d)", options.PodIndex, len(pods.Items))
+	}
+	pod := pods.Items[options.PodIndex]
+
+	if len(pod.Spec.Containers) == 0 {
+		return errors.New("pod has no containers")
+	}
+	container := pod.Spec.Containers[0]
```
vladimirvivien, line 379: Should `ExecInDeployment` also support specifying the container as well?

```go
func WithContainerIdxInDeployment(idx int) ExecInDeploymentOption { ... }
```

klient/k8s/resources/resources.go
vladimirvivien: A better suggestion that would make naming easier is to generalize the optional type so that it can be used in different context. Instead of using customized selectors, use existing mechanism for selecting pod and containers:

```go
type deploymentOption struct {
    listOption ListOption
    podName string
    podIdx int
    containerName string
    containerIdx
}
```
This change would allow the reuse of the existing `ListOption` type in this package along with other optional arguments.

```go
err := c.Client().Resources().ExecInDeployment(
    ctx, 
    c.Namespace(), 
    deployment.Name, 
    cmd, &stdout, &stderr, 
    resources.WithDeploymentLabelSelector("labelname=labelval"),
    resources.WithDeploymentFieldSelector("FieldName=FieldVal"),
    resources.WithDeploymentPodName("podName"),
     resources.WithDeploymentPodIdx(5),
     resources.WithDeploymentContainerName("podName"),
     resources.WithDeploymentContainerIdx(12),
)
```
This now is general enough to support all selection scenarios.

vladimirvivien:
@keeprocking I left some comments. Please take a look.

keeprocking:
Hi, sorry for the delay — got a lot going on in my life right now.

I'm working on the suggested changes, but they do require some time to get it done right.

klient/k8s/resources/resources.go
```diff
@@ -331,25 +331,94 @@ func (r *Resources) ExecInPod(ctx context.Context, namespaceName, podName, conta
 	return nil
 }
 
-type execInDeploymentOptions struct {
-	PodIndex int
+type matcher[T any] func([]T) (T, error)
+
+type predicate[T any] func(T) bool
```
vladimirvivien, line 336: 👍 on the use of type parameters here!!!

vladimirvivien (approved):
This is an awesome implementation.

PR #334, opened 2023-10-17
Do not run before after feature steps if feature skipped

matrus2:
/kind bug

Fixes #330

PR #517, opened 2025-08-16
fix: make test scope per-feature

ndbaker1:
#### What type of PR is this?

/kind bug

Fixes a regression introduced in https://github.com/kubernetes-sigs/e2e-framework/pull/334 when skipping a single feature inside a `testEnv.Run` call, all other features get skipped due to how the `t.Skip` was moved out of the per-feature execution test scope to address a separate concern with feature before/after steps.

This PR makes the entire feature processing call part of single test scope, that way feature skipping does not cause noise to neighbors.

Fixes #

```release-note
N/A
```

```docs
N/A
```


cartermckinnon:
/lgtm

@harshanarayana / @cpanato can one of you take a look at this? It's been annoying us for a while 😄

cpanato (approved):
lgtm

thanks

PR #537, opened 2025-12-04
feat: expose DeleteOptions in envfuncs.DeleteNamespace

cartermckinnon:
#### What type of PR is this?

/kind feature

I want to delete a test's namespace in the foreground, to ensure the namespace is not left `Terminating` after the test completes.

I've exposed the `resources.DeleteOptions` to make this possible using `resources.WithDeletePropagation`.

Fixes #

```release-note
envfuncs.DeleteNamespace now accepts resources.DeleteOptions.
```

```docs

```


vladimirvivien (commented):
This looks good @cartermckinnon . Since it is changing the signature, please add some tests.

PR #557, opened 2026-04-09
fix: Pass context to cluster providers when creating clusters

chancez:
#### What type of PR is this?

/kind feature
/kind deprecation

Currently if you cancel the context provided to `provider.Create` most providers ignore the context when running the command to create the cluster. Since cluster creation can be relatively long, it would be ideal if the providers actually forwarded this so the creation can be cancelled.

I considered updating existing functions but based on the discussion in #541 it would be preferred to make changes backwards compatible, so I've introduced new context aware functions and marked existing functions that do not takea  context as deprecated.

I also considered updating more code to pass context, but I figured let's keep the scope focused on the `Create()` calls where it arguably matters most.

```release-note
Update third_party cluster providers to propagate context.Context to commands when creating clusters
```

### DIFF
--- a/pkg/utils/command.go
+++ b/pkg/utils/command.go
@@ -111,8 +111,16 @@ func RunCommandContext(ctx context.Context, command string) *exec.Proc {
 
 // RunCommandWithSeperatedOutput run command and returns the results to the provided
 // stdout and stderr io.Writer.
+//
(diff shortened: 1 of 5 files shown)


cpanato (commented):
/ok-to-test

chancez:
@cpanato Tests should be passing now

cpanato (approved):
thanks!

